PR #6, opened 2021-08-20
Proof-read docs

aaronbates:
Feel free to reject/change any suggestions.

Other notes:

- It would make sense to update the repo description to "DataCamp's Design System"
- We should probably make all scripts the same format, for example we have `build:thing` and `test-unit` where I would expect to see `test:unit`.


blurbyte (approved):
Looks great 🥇 Also reverted back npm script names to the ones with colons (as suggested).

PR #32, opened 2022-05-11
Update ESLint Config for imports

ixTec:
# Update ESLint Config for sorting imports into groups

### ❓ Misc Changes

- Updated ESLint configuration to add the `eslint-plugin-import` plugin and associated config
- Fixed incorrect ordering of imports, per the new configuration rules

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,8 +42,8 @@ module.exports = {
         groups: [
           'external',
           'index',
-          'sibling',
           'parent',
+          'sibling',
           'internal',
           'builtin',
           'object',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,8 @@ module.exports = {
     'plugin:react/recommended',
     'plugin:@typescript-eslint/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:import/recommended',
+    'plugin:import/typescript',
     'prettier',
   ],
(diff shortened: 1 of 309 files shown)


blurbyte (changes requested):
Looks awesome! I'm not a big fan of 'on save' reformatting tbh. I think it would be better to add it to precommit hook as with prettier formatting. The config itself is in **.husky** directory, **pre-commit** file. Probably something along the lines `yarn eslint ./ --fix` should do the trick (run it before prettier).

ixTec:
@blurbyte I think precommit hooks for actual changes are a fairly big no-no (You end up with changes being made as part of your commit), but adding our linting run to a precommit hook without the `--fix` flag should be done for sure so we at least reject a commit with invalid changes / linting issues.

As for the on-save, I think it makes being able to adhere to these sort of rules much easier. I already have prettier configured locally to run on save, so could just set the onsave formatting in my local config too if you don't wish it to be project-wide

blurbyte:
> I think precommit hooks for actual changes are a fairly big no-no
People doing it for years with prettier and linters too and it's rather hand. The trick is to prettify / lint only staged files so it doesn't run through whole repo. For prettier we are using `pretty-quick` lib. There is probably something similar for eslint.
If it's hard to do it's fine to run `lint` npm script manually. I prefer to NOT have it for VSCode on save.

blurbyte (approved):

PR #33, opened 2022-05-13
Add new asset components and update LinkedIn colours

ixTec:
# Add new asset components and update linkedin colours

### 🚀 Major Changes

- Added new `Asset` components (e.g. `<BarChart3d />` - [WF-567](https://datacamp.atlassian.net/browse/WF-567)
    - Includes a variety of '3D' assets and an avatar placeholder

### 🔧 Minor Changes

- Updated colours for `<LinkedInBrand />` icon and the `brandLinkedIn` design token - [WF-583](https://datacamp.atlassian.net/browse/WF-583)

### ❓ Misc Changes

- Restructured the `icon` folder to match `asset`
- Moved several `tools` utilities into a shared location
- Docs site type resolution will now fallback to `kind` if no `value` is present
    - Solves issue with non-string union types (e.g. `string | number`)


doc-site/pages/components/asset.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+import BasicUsage from '../../examples/asset/basic-usage';
+
+export default Layout;
+
+# Asset
+
+🚧 This page is **under construction**.
+
+An asset is a visual element.
```
blurbyte, line 16: I would skip anything besides 'This page is under construction' for actual guidelines.

blurbyte (changes requested):
Love the idea of putting generated components in separate folder next to raw (maybe name _generated_ for this dir would be better than _output_?) and adding the comment at the top.

**About nesting folders**

Personally I would try to avoid nested folders as much as possible:

- Rather than putting 3d / other assets in separate folders I would suggest adding suffix to file name `bar-chart-3d`, etc. For other I guess it could be omitted. It would make file name - component name relationship very clear and straightforward.
- I would put all tools utils in `helpers.js` file. Probably not that many different kinds of utils atm to justify putting them in different folders.

**About letter casing**:

- Since we had to use  lowercase `3d` suffix I thin we should go with `Svg` in function names
- To make functions naming more consistent I think it makes sense to prefix them with action verb (`get`, etc.)

**About width and hight for assets**:
- I am not sure if it's best idea to let users set it explicitly. Imagine asset of default size of 240x180 set to 180x140 by the user. I think if we want to go that route we should maybe offer `scale` prop. But tbh I imagine in most of the cases they will be used as responsive images with `svg { width: 100%; }` CSS applied.

**About docs**
- Looks nice!
- I think on the grid assets could be a bit smaller (maybe fit 5 instead of 4 by default)
- For examples I would add 1 more showing how to use it in 'responsive way'. Or it may be basic usege?

**About tests**:
For stories maybe put them on the grid, render name of each asset (same as in docs) and make them a bit smaller.

blurbyte (approved):
Awesome! Let's merge.

PR #34, opened 2022-05-20
[WF-567] Branding Assets

ixTec:
# Added Branding Asset components, new regular assets and downloadable svg asset bundle

### 🚀 Major Changes

- Added new `Branding Asset` components (e.g. `<DataCampLogo />` - [WF-567](https://datacamp.atlassian.net/browse/WF-567)
    - Includes the official DataCamp logo and logomark, with tag variants.
    - Configurable with `inverted` or `monochrome`
- Technology Logos and Logomarks are now available as `Asset` components [WF-567]
- Partner Logos are now available as `Asset` components
- Added new `Minus` icon

### 🔧 Minor Changes

- Added downloadable `zip` to docs site, containing all available branding and regular assets in `SVG` format
    - Requires new dev dependency `archiver`

### ❓ Misc Changes

- Refactored svg generation into helper
- Alphabetically ordered all component categories on docs site

[WF-567]: https://datacamp.atlassian.net/browse/WF-567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blurbyte (changes requested):
Good job 👏  overall!

I think alphabetic ordering in the SideNav already makes it easier to use. New _tools_ folder structure makes sense.

**Asset**

Assets mostly looks good. I would maybe add information at the top of docs that they are _responsive_ and if width not height are NOT provided they will fit into the size of container (parent element). And maybe an info that they by default they inherit _color_ of their container too.

Docs:
- for each asset example please provide `color: ${tokens.colors.navy}` for wrapper, so they will look nicer (rather than pitch black)
- maybe increase padding between each asset a bit (maybe try `tokens.spacing.medium`), which should increase readability
- would split **Logos** to **Tech Logos** and **Partner Logos**
- **Alpa** section, would rename it to **ALPA Loop** (component names should stay the same)
- In Alpa section I would put them only in 2 columns, because they are expected to be rather large (so I guess we want it configurable)
- are we missing **WafflesLogomark** (non colored version)?
- for **Basic Usage** example I would make it smaller (~100px width), and at the same time show most common usage put it in div with some width set
- probably it would be nicer to have download link for each section, e.g. 'Download 3D Assets' secondary button under 3d assets section (so I would create `/downloads` dir in public folder and put all ZIPs there
- you can commit auto generated download ZIPs bundles

**Branding Asset**

Unfortunately I don't like this name. Would prefer simple **Logo**.

I also think what we have in Figma for tags is a bit better too:
- DataCampLogoTag -> DataCampFullTagline
- DataCampLogomarkTag -> DataCampTagline

As for component itself I think it is a bit over-engineered:
- our logos shouldn't use `currentColor` they are always using green / white / navy from tokens for whatever variant
- I would completely drop auto generation part (maybe keep SVGs for ZIPs?)
- Because monochrome / inverted is just simple color change you can creat simple mapping between flags and logomark shape and text shape

So for example for `DataCampLogo` it would be:
```ts
function DataCampLogo({
  inverted = false,
  monochrome = false,
  ...restProps
}: DataCampLogoProps) {
// calculate textColor and logomarkColor based on the inverted and monochrome flags
  return (
    <svg
    xmlns="http://www.w3.org/2000/svg"
    viewBox="0 0 173 36"
    xmlSpace="preserve"
    {...props}
  >
  // text path
    <path
      d="M42.56 27.099a5.694 5.694"
      fillRule="evenodd"
      clipRule="evenodd"
      fill={textColor}
    />
// logomark path
    <path
      d="M11.699 8.514v8.333L2.858"
      fillRule="evenodd"
      clipRule="evenodd"
      fill={logomarkColor}
    />
  </svg>
  );
}
```

**Docs**

List of all logos is missing headline, e.g. All Available DataCamp Logos.

Logos are too big in examples, 200px width or so will look better.

For imports in this case we ca show all of them (because there are only 4, compared to 30+ assets and 200+ icons).

src/brand/brand.types.tsx
```diff
@@ -0,0 +1,6 @@
+export type BrandGenericProps = {
+  /* Whether the brand is inverted in color */
```
blurbyte, line 2: Not sure if it grammatically makes sense here, but everywhere else there is period at the end of description.

doc-site/components/all-asset-types.tsx
```diff
@@ -3,10 +3,25 @@ import * as allAssets from '@datacamp/waffles/asset';
 
 import AssetGrid from './asset-grid';
 
+const groupMaxColumnCount = {
```
blurbyte, line 6: Really like this approach.

blurbyte (changes requested):
There are a few minor things here and there to fix, but overall looks great 🥇 .

Basic usage example for asset is too big:
Screenshot 2022-05-26 at 10 38 54

Assets examples should be wrapped with div with `color: ${tokens.colors.navy}`.

These ALPA Loops are off too, I think all should be grey. Maybe it would make sense to render them agains navy bgr, since there are white arrows (but skip it if it's a lot of work):
loops

doc-site/pages/components/brand.mdx
```diff
@@ -0,0 +1,72 @@
...
+</Example>
+
+## Props
+
+### Brand
+
+In addition to `inverted` and `monochrome` listed below, Brand components accept all props a regular `SVGElement` would, such as `width`, `height`, `aria-` and `data-` attributes.
+
+<PropsTable
+  metadata={require('!!extract-react-types-loader!../../../src/brand/datacamp-logo')}
```
blurbyte, line 57: Should be `data-camp-logo`.

doc-site/components/all-asset-types.tsx
```diff
@@ -3,10 +3,25 @@ import * as allAssets from '@datacamp/waffles/asset';
...
+  'ALPA Loop': 2,
+  '3D': 4,
+  Logomark: 3,
```
blurbyte, line 9: I would increase it to 4, they are too big now :)

doc-site/examples/asset/custom-size-color.tsx
```diff
@@ -0,0 +1,17 @@
+import { DoubleArrow3d } from '@datacamp/waffles/asset';
+import { tokens } from '@datacamp/waffles/tokens';
+import { css } from '@emotion/react';
+
+function Example() {
+  return (
+    <div
+      css={css`
+        color: ${tokens.colors.grey};
```
blurbyte, line 9: Lets use `greyDark` here. Regular `grey` is too light in this case.

blurbyte (changes requested):
We are very close :)

Sorry for changing my mind again (I promise it's the last time), but for downloading, we can add a bar and a button (with nice download icon) same as for show code for example:
bar

In brand section there is still old link.

PR #35, opened 2022-05-25
[WF-582] Update sizing for buttons, inputs and formfields

ixTec:
# Update sizing for buttons, inputs and formfields

### 🚀 Major Changes

- Added new `fontSizes.xsmall` token with `10px` value
- `sizing.small` token now has a value of `28px` (was `24px`)

### 🔧 Minor Changes

- `small` variant `Button`, `FormField`, `Select`, `TextArea` and `Input` components now use  `fontSizes.small` (was `fontSizes.medium`)
    - These variants will also now use `small` icons by default (was `medium`)
    - `medium` and `large` variants are unaffected
- Components with an _icon-like_ property (e.g. `icon`, `iconLeft` or `iconRight`) have had their intrinsic type updated from `React.ReactNode` to `JSX.Element`

### ❓ Misc Changes

- Removed small variant buttons and inputs from roadmap


src/button/button-internal.tsx
```diff
@@ -60,6 +60,16 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
 
   const { focusProps, isFocusVisible } = useFocusRing();
 
+  function getClonedIcon(originalIcon: React.ReactElement) {
+    // Check if the icon has a provided custom size prop already
+    return originalIcon.props['size']
```
blurbyte, line 65: I think dot could be used safely here.

src/button/button-internal.tsx
```diff
@@ -60,6 +60,16 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
+      ? originalIcon
+      : cloneElement(originalIcon, {
+          // Handle large buttons having medium sized icons by default
```
blurbyte, line 68: Maybe mentioned medium and small sizes being passed as it is.

src/input/input-internal.tsx
```diff
@@ -52,15 +52,39 @@ function InputInternal(
...
+            data-testid="search-icon"
+            size={size === 'large' ? 'medium' : size}
+          />
         </IconLeft>
       );
     } else if (iconLeft) {
-      return <IconLeft {...{ size, inverted }}>{iconLeft}</IconLeft>;
+      return (
+        <IconLeft {...{ size, inverted }}>
+          {cloneIconElement(iconLeft as React.ReactElement)}
```
blurbyte, line 85: I think in similar way to `Button` we can safely change `iconLeft` and `enhancerRight` type to `JSX.Element`, since realistically we want only 1 element to be passed.

src/input/input-internal.tsx
```diff
@@ -84,7 +109,7 @@ function InputInternal(
         <EnhancerRight {...{ size, inverted }}>
           {Children.map(enhancerRight, (child) => {
```
blurbyte, line 110: With `enhancerRight` both _Children.map_ nor _isValidElement_ won't be required anymore.

src/input/input-internal.tsx
```diff
@@ -52,15 +52,39 @@ function InputInternal(
     setIsFocused(false);
   }
 
+  function cloneIconElement(
```
blurbyte, line 55: Tbh I think in this case duplication is fine (it's only 1 line of code if put inside `render` functions), so no need for this utility function. You can just assign `const iconSize = size === 'large' ? 'medium' : size` somewhere at the top of component.

src/form-field/form-field.tsx
```diff
@@ -50,11 +53,12 @@ function FormField({
       'aria-errormessage': errorId,
       'aria-describedby': errorId,
     }),
+    size: size,
```
blurbyte, line 56: No need to write `size: size`, `size` alone is enough.

src/input/input-internal.tsx
ixTec: This is used in three places within `input-internal` and handles the different use cases. Using a declaration of `iconSize` won't work as the icons could (in theory) have different sizes

src/button/button-internal.tsx
ixTec: `.props` isn't an explicitly typed object for the icon, so we'd have to tell typescript to ignore it, which should be avoided where possible

src/input/input-internal.tsx
blurbyte: I was thinking: since it's just 1 liner for iconLeft and enhancerRight `iconLeft.props.size ? cloneElement(iconLeft, { inverted }) : cloneElement(iconLeft, { inverted, size: iconSize })` it might be overkill to provided it via separate function.

ixTec: I think for code readability, it makes sense to keep this as a separate function here

blurbyte (changes requested):
We are pretty close. A few minor things:

- Probably it's good idea to add snapshot of each size for `FormField`
- I would test all sizes in `stories/form-field-label`

Also I think it would make sense to add `small` TextArea but we need design spec for it from Petra first.

blurbyte (approved):
Looks good! 

Just a few really minor things:

- for that bookmarkStyle I think `vertical-align: baseline` works the best
- casing in FormField sizes example for labels seems to be off (title casing, tbh I'm not 100% sure which is correct, probably we are missing guidelines for that, but right now regular sentence case is used everywhere else)
- I think side padding in TextArea small should be smaller to be consistent with small input, ~but waiting for this one for Petra to be confirmed (we can update it in some PR later)~

Be free to merge otherwise 🚀

PR #36, opened 2022-05-25
[WF-546] Menu

blurbyte:
# New Menu dropdown component

### 🚀 Major Changes

- Added new `Menu` component
    - It has regular and inverted version
    - Placement is configurable
    - Ships with accessible `Menu.Item` and `Menu.Button` subcomponents; Radio and Checkbox planned for future release
    - `useMenu` hook exposes some internal `Menu` data
    - ⚠️ There is issue with a11y when custom styles are provided with `css` prop to `Menu` subcomponents; most likely will address it with making it required to wrap these subcomponents with `Menu.Category` (which is mentioned in docs already)
    - `SideNavigation.Item` is very similar to `Menu.Item`; probably should have `label` instead of `children`, but will tackle it next release

### 🔧 Minor Changes

- Added `xxsmall` (10px) font size design token
- Added `deepChildrenMap` helper
- Updated `Tooltip` internals
- Updated `Portal` component and how it is used internally, so it works better with refs
- Updated the size of `SideNavigation` category label on desktop to use new font size token
- Tighten up some TS typings

### ❓ Misc Changes

- Updated roadmap with what we discussed in Belgium
- Bumped some dependencies

## Screenshot(s)

Regular menu with all features enabled:
Screenshot 2022-05-25 at 13 39 56

Inverted menu:
Screenshot 2022-05-25 at 13 40 44


src/menu/category.tsx
```diff
@@ -0,0 +1,44 @@
...
+  /* Determines if a subtle horizontal divider is shown above the category label. */
+  noDivider?: boolean;
+  /* A list of `Menu` subcomponents, such as `Menu.Item` or `Menu.Button`. */
+  children: React.ReactNode;
+} & React.HTMLAttributes<HTMLDivElement>;
+
+function Category({
+  label,
+  children,
+  noDivider = false,
```
ixTec, line 21: I think we should change this to `hasDivider` and default to false. Right now, if you have a category as your first child under `Menu`, you have to manually disable the divider for it to look right in a default configuration.

For further siblings, having a label at the very least looks visually okay and so a horizontal divider only further improves it there but I think having it as opt-in rather than opt-out would be a better approach

src/menu/item.tsx
```diff
@@ -0,0 +1,90 @@
...
+  /* [skip docs] */
+  index?: number;
+  /* Optional description shown below the label. */
+  description?: React.ReactNode;
+  /* An icon displayed to the left. Could be any [icon](/components/icon) from Waffles (use default `medium` size) or a custom component. */
+  iconLeft?: React.ReactNode;
+  /* An icon displayed to the right. Could be any [icon](/components/icon) from Waffles (use default `medium` size) or a custom component. */
+  iconRight?: React.ReactNode;
+  /* Displays small notification dot next to the label. Useful for highlighting important items. */
+  showAlert?: boolean;
```
ixTec, line 29: I think `showNotificationDot` or `hasNotificationDot` is a better term for this prop

ixTec (changes requested):

doc-site/pages/components/menu.mdx
```diff
@@ -0,0 +1,94 @@
...
+<PropsTable
+  metadata={require('!!extract-react-types-loader!../../../src/menu/item')}
+  isPolymorphic
+/>
+
+### Menu.Button
+
+Accepts all props Waffles [Button](/components/button/#props) would.
+
+### useMenu
```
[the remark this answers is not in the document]
blurbyte: Yeah, something to think about. Because for hooks you can:
1. Provide config for hook itself
2. It also returns some data
For now I would leave as it is, since this particular hook is not that critical (aka is 100% optional to use, not like in e.g. Toast where it is core functionality). We can improve it along with Petra's docs reimagination.

src/menu/item.tsx
blurbyte: Makes sense, I will change it to `showNotificationDot`.

src/menu/category.tsx
blurbyte: Something to think about. I used the same prop as in `SideNavigation.Category`. So would make sense to update it there as well. Also I think it would make sense to change how `SideNavigation.Item` content is provided (right now via `children` but probably makes sense to do it via `label` prop). That's 2 breaking changes. Let's push it to the next release.

src/menu/notification-dot.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import { alertDotStyle } from './styles';
```
ixTec, line 3: `alertDotStyle` should be renamed to match `notificationDot`

src/menu/styles.ts
```diff
@@ -0,0 +1,220 @@
...
+    margin-left: ${tokens.spacing.small};
+    margin-right: ${tokens.spacing.small};
+
+    &:not(:last-child) {
+      margin-bottom: ${tokens.spacing.small};
+    }
+  `;
+}
+
+export function alertDotStyle() {
```
ixTec, line 209: `alertDotStyle` should be renamed to match `notificationDot`

ixTec (approved):

PR #40, opened 2022-05-30
[WF-600] Loader Component

ixTec:
# Loader component

### 🚀 Major Changes

- Add new `Loader` component
    - Configurable with `inverted` prop

## Screenshot(s):
Screenshot 2022-05-30 at 16 38 43

### DIFF
--- a/doc-site/pages/components/loader.mdx
+++ b/doc-site/pages/components/loader.mdx
@@ -1,0 +1,49 @@
+import * as LoaderExports from '@datacamp/waffles/loader';
+
+import Layout from '../../components/content-layout';
+import Example from '../../components/example';
+import Playground from '../../components/playground';
+import PropsTable from '../../components/props-table';
+import Imports from '../../components/imports';
+
+import BasicUsage from '../../examples/loader/basic-usage';
+import CustomSize from '../../examples/loader/custom-size';
+import Inverted from '../../examples/loader/inverted';
+
+export default Layout;
+
+# Loader
+
+An animated component to represent loading.
+
+🚧 This page is **under construction**.
+
+## Examples
+
+<Example title="Basic Usage" path="loader/basic-usage">
+  <BasicUsage />
+</Example>
+
+<Example title="Inverted" path="loader/inverted" darkPreview>
+  <Inverted />
+</Example>
+
+<Example title="Custom Size" path="loader/custom-size">
+  <CustomSize />
+</Example>
+
+## Props
+
+### Loader
+
+In addition to all custom props listed below, `Loader` accepts all props regular `HTMLDiv` would, such as `aria-labelledby`, `aria-describedby`, `data-` attributes.
+
+<PropsTable
+  metadata={require('!!extract-react-types-loader!../../../src/loader/loader')}
+/>
+
+## Imports
+
+You can import following components or utilities from this module:
+
+<Imports module={LoaderExports} path="@datacamp/waffles/loader" />
--- a/doc-site/pages/components/loader.mdx
+++ b/doc-site/pages/components/loader.mdx
@@ -36,7 +36,7 @@ An animated component to represent loading.
 
 ### Loader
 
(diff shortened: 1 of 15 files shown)


doc-site/pages/components/loader.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+<Example title="Custom Size" path="loader/custom-size">
+  <CustomSize />
+</Example>
+
+## Props
+
+### Loader
+
+In addition to all custom props listed below, `Loader` accepts all props regular `HTMLDiv` would, such as `aria-labelledby`, `aria-describedby`, `data-` attributes.
```
blurbyte, line 39: I think it should be `HTMLSvg`.

src/loader/loader.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { tokens } from '../tokens';
+
+import { svgStyle, svgWrapperStyle, wrapperStyle } from './styles';
+
+type LoaderProps = {
+  /* Whether the loader is inverted in color or not. */
+  inverted?: boolean;
+} & React.SVGAttributes<SVGElement>;
+
+const Loader = ({
```
blurbyte, line 12: Please use regular `function()`. No need to specify return type as well.

src/loader/loader.tsx
```diff
@@ -0,0 +1,38 @@
...
+  inverted = false,
+  ...restProps
+}: LoaderProps): React.ReactElement => {
+  const { height, width } = restProps;
+  return (
```
blurbyte, line 17: I would put new line here.

src/loader/styles.ts
```diff
@@ -0,0 +1,30 @@
+import { css } from '@emotion/react';
+
+import { loaderEndMask, loaderStartMask, stroke } from './keyframes';
+
+const animationSettings = '2s infinite alternate';
+
+export function wrapperStyle(height: string, width: string) {
```
blurbyte, line 7: As per convention it should be `wrapperStyle({ height, width })` as options.

src/loader/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+    display: block;
+    height: ${height};
+    width: ${width};
+    animation: ${loaderStartMask} cubic-bezier(0.65, 0, 0.55, 1)
+      ${animationSettings};
+    will-change: clip-path;
+  `;
+}
+
+export const svgWrapperStyle = css`
```
blurbyte, line 18: Maybe `loaderWrapperStyle`?

src/loader/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+  animation: ${loaderEndMask} cubic-bezier(0, 0, 0.85, 1) ${animationSettings};
+  will-change: clip-path;
+`;
+
+export const svgStyle = css`
```
blurbyte, line 23: Maybe `loaderStyle`?

src/loader/loader.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <div
+      css={wrapperStyle(height as string, width as string)}
+      data-testid="loader-wrapper"
+    >
+      <div css={svgWrapperStyle}>
+        <svg viewBox="0 0 2640 3444" css={svgStyle} {...restProps}>
+          <title>Loading</title>
```
blurbyte, line 24: Because of SEO it is discouraged to use `<title>` anywhere besides HTML body head. I think loader doesn't need to have any a11y attributes. I would even put `aria-hidden="true"` there. Imagine button with label `Loading...`: it will be announced 2 times.

blurbyte (changes requested):
Good job. Just a few minor changes.

doc-site/pages/components/loader.mdx
ixTec: Updated to be `SVGElement`

blurbyte (approved):
Let's merge 🚀

PR #41, opened 2022-05-31
[WF-598] SideNavigation and Drawer adjustments

blurbyte:
# SideNavigation and Drawer width adjustments on mobile, iOS height fix for SideNavigation

### 🚀 Major Changes

- Updated the way how _width_ is calculated for `SideNavigation` and `Drawer` on mobile. Previously it was fixed 300px width, but now it is screen width with a small gap left to the right of panel (with max width being 356px). Everything is done with CSS.
- Added `xxlarge` spacing token.

### 🔧 Minor Changes

- Fixed the issues with `SideNavigation` height on iOS, when some portion of the bottom part of menu was cut. The solution is a bit creative: there are 2 fixed positioned panels animated in sync, one is background with `100vh` height and second one has position `bottom` set to 0. That way content is always visible and above browser bar on iOS. With new measurements units (such as `dvh`) it will be possible to simplify it, but right now they are not widely supported (basically work only in Safari iOS15+).

_Tested in Chrome & Safari on various iPhones, tested in Chrome on Android, tested in Chrome & Edge on Windows._

## Screenshot(s):

Updated width of SideNavigation and Drawer:
Recording: https://user-images.githubusercontent.com/20565536/171159603-039d2c26-d786-43bd-80a3-5bca17508f6d.mp4

On iPhone 11 (iOS13):
iphone11-ios13

On Samsung Galaxy9:
samsung-galaxy9

On iPhone 12 (iOS15):
ios-fixed

### DIFF
--- a/src/side-navigation/animated-sidebar.tsx
+++ b/src/side-navigation/animated-sidebar.tsx
@@ -3,14 +3,15 @@ import React from 'react';
 import { Portal } from '../portal';
 import { useAnimateTransition } from '../hooks';
 
-import { animatedSidebarStyle } from './styles';
+import { animatedSidebarStyle, animatedSidebarContentStyle } from './styles';
 import { useSidebar } from './sidebar-context';
 import Overlay from './overlay';
 import CloseButton from './close-button';
 
 type AnimatedSidebarProps = React.HTMLAttributes<HTMLDivElement>;
 
-function AnimatedSidebar(props: AnimatedSidebarProps) {
+// Incorporate two animated containers to make it work nicely on iOS
+function AnimatedSidebar({ children, ...restProps }: AnimatedSidebarProps) {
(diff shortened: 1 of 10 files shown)


src/side-navigation/animated-sidebar.tsx
```diff
@@ -20,7 +21,15 @@ function AnimatedSidebar(props: AnimatedSidebarProps) {
         <>
           <Overlay />
           <CloseButton />
-          <div {...props} css={animatedSidebarStyle({ isVisible: isOpen })} />
+          {/* Background sidebar which spans whole device heigh */}
```
ixTec, line 24: should be `height`

ixTec (changes requested):
just a few minor changes to comments / variable naming but otherwise looks great

ixTec (approved):

PR #46, opened 2022-06-06
[WF-597] Scrollable Tabs on all devices

blurbyte:
# Scrollable Tabs on all devices

### 🔧 Minor Changes

- Make `Tabs` scrollable on all device sizes (previously it was only on mobile). The gradient mask is shown when all tabs combined width is longer than their container. In future for desktop most likely we will implement either scroll arrows or show more button.
- Update focus style of `Tabs.Tab` to be more compact because of changes required to container overflow CSS property.

## Screenshot(s):

Large desktop:
Screenshot 2022-06-06 at 16 27 40

Tablet & small desktop:
Screenshot 2022-06-06 at 16 28 05

Mobile: 
Screenshot 2022-06-06 at 16 28 16

Updated focus style:
Screenshot 2022-06-06 at 16 28 33

### DIFF
--- a/src/tabs/__tests__/__snapshots__/tabs.spec.tsx.snap
+++ b/src/tabs/__tests__/__snapshots__/tabs.spec.tsx.snap
@@ -18,15 +18,6 @@ exports[`Tabs renders snapshot 1`] = `
   height: 0;
 }
 
-@media screen and (min-width: 480px) {
-  .emotion-0 {
-    overflow-y: visible;
-    overflow-x: visible;
-    -webkit-mask-image: none;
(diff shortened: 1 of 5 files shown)


ixTec (approved):
Overall looks great! Approved.

Only thought I have is about having a gradient on the left side of the tabs too, when you're scrolled across? Right now there's no visual representation that the tabs can be scrolled back and the harsh cut off when mid-tab looks a little odd

blurbyte:
Yes, we can think about it, although probably with bigger update related to improved desktop experience (although it's rather low prio).

PR #47, opened 2022-06-07
[WF-604] Menu indexing improvements 

blurbyte:
# Insert Description Here

### 🔧 Minor Changes

- Updated how `Menu` items are indexed (which handles focusing of items when navigating with keyboard), therefore it's now possible to apply custom styles to `Menu.Item` or use completely custom one. To make it possible, menu items _must_ be wrapped with `Menu.Category` (but docs already mentioned it, and all examples were implemented that way).

### ❓ Misc Changes

- Bumped `floating-ui`, and `react-aria` dependencies

## Screenshot(s):

https://user-images.githubusercontent.com/20565536/172416350-1a584a16-91de-4b82-b159-5240e3384cb1.mp4

### DIFF
--- a/doc-site/examples/menu/advanced-usage.tsx
+++ b/doc-site/examples/menu/advanced-usage.tsx
@@ -1,4 +1,6 @@
 import { forwardRef, useState } from 'react';
+import { css } from '@emotion/react';
+import { tokens } from '@datacamp/waffles/tokens';
 import { Menu, useMenu } from '@datacamp/waffles/menu';
 import {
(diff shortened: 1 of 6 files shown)


ixTec (approved):
Changes look great! Only comment I have is that it could be useful to make the "To make it possible, menu items must be wrapped with Menu.Category" as a more clear warning / important notice.

The menu children docs also state "Content of the menu, either Menu.Category or custom components." but I feel this should always be category from these new changes

PR #49, opened 2022-06-09
[WF-593] Avatar Component

ixTec:
# New Avatar component

### 🚀 Major Changes

- New `Avatar` component
    - Configurable with `variant` and `size` properties
        - `variant` colours match `Badge` component
    - Has `contentFill` prop to control whether the content fills the avatar space or restricts itself to the predefined spacing at each size

### ❓ Misc Changes

- Updated how the docs-site resolves union types containing generics

## Screenshot(s):

Screenshot 2022-06-09 at 12 40 33
Screenshot 2022-06-09 at 12 41 00
Screenshot 2022-06-09 at 12 41 14


src/avatar/avatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+} & React.HTMLAttributes<HTMLDivElement>;
+
+function Avatar({
+  size = 'medium',
+  variant = 'navy',
+  content,
+  contentFill = false,
+  ...restProps
+}: AvatarProps) {
+  if (typeof content === 'string' && content.length > 1) {
```
blurbyte, line 43: Tbh to make it resilient I would just take the first letter of a string and capitalize it (so warning won't be necessary). Of course in docs we still mention that ideally 1 character should be provided.

doc-site/pages/components/avatar.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+<Example title="Variants" path="avatar/variants">
+  <Variants />
+</Example>
+
+<Example title="Sizes" path="avatar/sizes">
+  <Sizes />
+</Example>
+
+<Example title="Fill Content Space" path="avatar/content-fill">
```
blurbyte, line 35: Maybe 'Fill Whole Avatar Space'?

src/avatar/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+    backgroundColor: tokens.colors.greyLight,
+  },
+};
+
+type AvatarWrapperProps = {
+  size: NonNullable<React.ComponentProps<typeof Avatar>['size']>;
+  variant: NonNullable<React.ComponentProps<typeof Avatar>['variant']>;
+};
+
+export function avatarWrapperStyle({ size, variant }: AvatarWrapperProps) {
```
blurbyte, line 104: I think it could be called `avatarStyle`.

src/avatar/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+    overflow: hidden;
+  `;
+}
+
+type AvatarContentWrapperProps = {
+  size: NonNullable<React.ComponentProps<typeof Avatar>['size']>;
+  contentFill: boolean;
+};
+
+export function avatarContentWrapperStyle({
```
blurbyte, line 123: Maybe `avatarContent`?

src/avatar/styles.ts
```diff
@@ -0,0 +1,146 @@
...
+    font-size: ${sizeMap[size].fontSize};
+    line-height: ${sizeMap[size].sizing};
+    width: ${sizeMap[size].sizing};
+    height: ${sizeMap[size].sizing};
+    font-family: ${tokens.fontFamilies.sansSerif};
+    overflow: hidden;
+  `;
+}
+
+type AvatarContentWrapperProps = {
```
blurbyte, line 118: As per convention `...StyleOptions`.

blurbyte (changes requested):
Amazing job. API seems very simple and straightforward to use (yet quite flexible) 🥇.

I would just introduce a little change to make it more resilient: when it's a string just take first character.

## For docs:

- Would rename _string_ to _text_ (string feels too technical)
- I think only one content type is enough here (and remove string content)
- For each subsection use `<Heading size="medium">`
- Would rename `profile-pic-placeholder.png` to `profile-pic-example.png`

Also noticed on mobile it's a bit broken:
Screenshot 2022-06-09 at 16 28 23
Probably setting `flex-shrink: 0` on wrappers should be enough (and `flex-wrap: wrap` for examples wrapper).

## For tests

### Unit
There is a way to handle static assets in jest: https://stephencharlesweiss.com/jest-handling-static-assets-mocks. That way you don't have to leave src empty.

In unit test snapshots I would mock `DataCampBrand` since it is polluting snapshots with bulk of redundant info:
```
jest.mock('../../path-to-data-camp-brand', () => 'DataCampBrand')
```
(or something similar, might be necessary to wrap it with div or sth).

### Stories

In cypress tests image is broken (probably have to put example PNG in public of workbench, since under the hood workbench is used to run these tests).

Would add some `gap` between each component so they don't touch each other.

For variant again I think it is enough to use only one content type.

Would write separate story for each content type (Text, Icon, Asset & Image).

doc-site/examples/avatar/content-fill.tsx
```diff
@@ -0,0 +1,133 @@
+import { css } from '@emotion/react';
+import { tokens } from '@datacamp/waffles/tokens';
+import { Heading } from '@datacamp/waffles/heading';
+import { Avatar } from '@datacamp/waffles/avatar';
+import { AvatarPlaceholderColored } from '@datacamp/waffles/asset';
+
+function Example() {
+  return (
+    <>
+      <Heading as="h4" size="medium">
```
blurbyte, line 10: You don't have to specify `as` here, `medium` defaults to `h4`.

doc-site/examples/avatar/sizes.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { GitLogomark } from '@datacamp/waffles/asset';
+
+function Example() {
+  return (
+    <>
+      <Heading as="h4" size="medium">
+        Text Content
+      </Heading>
+      <div
+        css={css`
```
blurbyte, line 15: Maybe define `wrapperStyle` at the top. It's used 3 times (and clear margin in the last one).

doc-site/pages/overview/roadmap.mdx
```diff
@@ -17,7 +17,6 @@ Planned features are listed in priority order.
 | Feature                     | Description                                                                                           | Delivery         |
 | --------------------------- | ----------------------------------------------------------------------------------------------------- | ---------------- |
 | EmptyState                  | Component to show when there is nothing to show.                                                      | Q2&nbsp;2022     |
-| Avatar                      | Standarized way to display images in a circle.                                                        | Q2&nbsp;2022     |
```
blurbyte, line 20: I think we can remove question mark from 'Making it Work for Marketing' since it will happen next Q for sure.

doc-site/pages/overview/roadmap.mdx
```diff
@@ -28,15 +27,20 @@ Planned features are listed in priority order.
 ## Component Status
 
 <ComponentStatusTable>
+  <ComponentStatus
```
blurbyte, line 30: Let's add `Card` and `ResizableElement` to the list, both as unavailable.

jest.config.js
```diff
@@ -14,4 +14,8 @@ module.exports = {
   modulePathIgnorePatterns: ['<rootDir>/dist/'],
   testRegex: '(.|-)spec\\.(js|ts|tsx|jsx)$',
   snapshotSerializers: ['@emotion/jest/serializer'],
+  moduleNameMapper: {
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      require.resolve('./jest/image-file-mock.js'),
```
blurbyte, line 19: Maybe let's create `test-setup` dir in the root, and put there `test-setup.ts` (could be `setup` or `initial-setup`) and this file there.

src/avatar/__tests__/avatar.spec.tsx
```diff
@@ -0,0 +1,92 @@
+import { render } from '@testing-library/react';
```
blurbyte, line 1: I think the use case where some longer string is provided and the only first char is left should be tested.

src/avatar/avatar.tsx
```diff
@@ -0,0 +1,54 @@
...
+  variant = 'navy',
+  content,
+  contentFill = false,
+  ...restProps
+}: AvatarProps) {
+  return (
+    <div css={avatarStyle({ size, variant })} {...restProps}>
+      <div css={avatarContentStyle({ size, contentFill })}>
+        {content && typeof content === 'string'
+          ? (content as string).charAt(0).toUpperCase()
```
blurbyte, line 47: TS is smart enough to figure it out, no need to cast it.

src/avatar/__tests__/avatar.spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const avatarImage = getByAltText('Avatar test image');
+    expect(avatarImage).toBeInTheDocument();
+  });
+
+  describe('renders snapshot of', () => {
+    variants.forEach((variant) => {
+      sizes.forEach((size) => {
+        it(`variant ${variant} and size ${size}`, () => {
+          const { container } = render(
+            <>
```
blurbyte, line 70: It should be wrapped in `<div>`. When you're doing `container.firstChild` it grabs only `Avatar` with text content.

src/avatar/__tests__/avatar.spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+              <Avatar variant={variant} size={size} content="A" />
+              <Avatar
+                variant={variant}
+                size={size}
+                content={<DataCampBrand />}
+              />
+              <Avatar variant={variant} size={size} content={<GitLogomark />} />
```
blurbyte, line 77: Do we need this one? If yes please mock it in similar way to DCBrand.

blurbyte (changes requested):
Great job. Just a few small tweaks.
Just noticed texts size in status table of 'Unavialable' should be updated after button change:
Screenshot 2022-06-10 at 13 03 39

jest.config.js
ixTec: Little confused by this comment - can you expand please?

PR #50, opened 2022-06-09
[WF-580] Add new assignment icon assets

ixTec:
# New Assignment Assets

### 🚀 Major Changes

- Added new assignment assets

### ❓ Misc Changes

- Fixed issue where `generate:assets` and `generate:icons` scripts weren't correctly removing older generated files before running script

## Screenshot(s):
Screenshot 2022-06-09 at 14 17 41


package.json
```diff
@@ -12,8 +12,8 @@
     "prebuild": "rimraf dist && yarn generate:all",
     "generate:all": "yarn generate:design-tokens && yarn generate:assets && yarn generate:icons",
     "generate:design-tokens": "rimraf src/tokens/tokens.ts && node ./tools/generate/generate-design-tokens.js",
```
blurbyte, line 14: Maybe wrap it with escaped quotation marks too, just to be safe (I think I tested it on Windows but not 100% sure).

blurbyte (approved):
Looks good. Any idea how to fix the Asset docs ordering of sections, since it's different on local dev & prod.

PR #55, opened 2022-06-23
[WF-538] Empty State Component

ixTec:
# Empty State Component

### 🚀 Major Changes

- Added new `EmptyState` component
    - Has the following children:  `EmptyState.List`, `EmptyState.ListItem` and `EmptyState.Content.
    - Can set row or column layout via `direction` property
    - Can be inverted using `inverted` property
    - Heading is set via `title` property

## Screenshot(s):

Screenshot 2022-06-23 at 15 24 46
Screenshot 2022-06-23 at 15 25 11


ixTec:
**Things to consider:**

- Using name `direction` for property to be similar to the vanilla `flex-direction` css property
- We may want `title` prop to be type `JSX.Element` instead of `string`, to support configuration of level?

blurbyte:
I wold change `direction` to orientation (vertical / horizontal).
I think wouldn't hurt if `title` will be `React.ReactNode`.
Don't think we need `EmptyState.Content` wrapper.

blurbyte (changes requested):
Looks really good 🥇.

A few random comments (apart what is inline):

1. I would definitely NOT use 'Lorem ipsum...' in examples (for tests it's fine). Some of the samples from Petra's inventory research could be used here.

2. Also we may show an example (don't have to separate section, just change one of existing ones):
    - without `List` at all
    - `ListItem`s without icons

3. Looks like TS stop picking up `src` dir after my last jest config adjustments. I think we have to update `include` field in root `tsconfig.json` to ` "include": ["src", "test-setup/initial-setup.ts"]`.

src/empty-state/list-item.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { listItemStyle } from './styles';
+
+type ListItemProps = {
+  /* An icon displayed to the left of the list item. Could be any [icon](/components/icon) from Waffles (use default `medium` size) or a custom component. */
+  icon?: JSX.Element;
+  /* The content of the list item. Most of the time should be plain text. */
+  children: React.ReactNode;
+};
+
+function ListItem({ icon, children, ...restProps }: ListItemProps) {
```
blurbyte, line 12: I would use `<Text as="li">` under the hood since it bakes in a lot of font-specific styling.

src/empty-state/empty-state-internal.tsx
```diff
@@ -0,0 +1,76 @@
...
+  imageStyle,
+} from './styles';
+
+type EmptyStateBaseProps = {
+  /* Image content element. In general pass Waffles [Asset](/components/asset). */
+  image?: JSX.Element;
+  /* Content of the body, can include `EmptyState.List` and any Waffles or custom component. */
+  children?: React.ReactNode;
+  /* Title heading for the content. */
+  title?: React.ReactNode;
```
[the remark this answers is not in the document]
ixTec: Specified as optional because Petra wanted it configurable, since sometimes a title won't be provided

src/empty-state/empty-state-internal.tsx
```diff
@@ -0,0 +1,76 @@
...
+import {
+  contentWrapperStyle,
+  emptyStateStyle,
+  headingStyle,
+  imageStyle,
+} from './styles';
+
+type EmptyStateBaseProps = {
+  /* Image content element. In general pass Waffles [Asset](/components/asset). */
+  image?: JSX.Element;
```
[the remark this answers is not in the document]
ixTec: Specified as optional because Petra wanted it configurable, since sometimes an image won't be provided in all use cases

src/empty-state/empty-state-internal.tsx
```diff
@@ -0,0 +1,76 @@
...
+          : orientation || 'horizontal',
+        isCentered: !!isCentered,
+        inverted,
+      })}
+      {...restProps}
+    >
+      {image && <div css={imageStyle}>{image}</div>}
+      <div css={contentWrapperStyle}>
+        {title && (
+          <Heading css={headingStyle} size={'xlarge'} inverted={inverted}>
```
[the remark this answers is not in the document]
ixTec: This is already `xlarge` ?

src/empty-state/list-item.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { listItemStyle } from './styles';
+
+type ListItemProps = {
+  /* An icon displayed to the left of the list item. Could be any [icon](/components/icon) from Waffles (use default `medium` size) or a custom component. */
+  icon?: JSX.Element;
```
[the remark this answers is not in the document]
ixTec: I think sticking to more generalised terms for props is better. This is _the icon_ for the list item, and makes no sense to use `leftIcon` without a matching `rightIcon`

src/empty-state/styles.ts
```diff
@@ -0,0 +1,106 @@
+import { css } from '@emotion/react';
+
+import { EmptyState } from '.';
+
+import { tokens } from '../tokens';
+import { hexToRgba } from '../helpers';
+
+function emptyStateBaseStyle() {
```
blurbyte, line 8: This one actually don't have to be func (only used internally) so it can be used later as `${emptyStateBaseStyle}`.

src/empty-state/styles.ts
```diff
@@ -0,0 +1,106 @@
...
+    flex-shrink: 0;
+    width: 100px;
+    height: 100px;
+  `;
+}
+
+export function contentWrapperStyle() {
+  return css`
+    display: flex;
+    flex-direction: column;
```
blurbyte, line 69: These 2 lines could be omitted, so button won't span full width of container, which is a bit too heavy visually:
Screenshot 2022-07-07 at 12 33 23

src/empty-state/styles.ts
```diff
@@ -0,0 +1,106 @@
...
+import { EmptyState } from '.';
+
+import { tokens } from '../tokens';
+import { hexToRgba } from '../helpers';
+
+function emptyStateBaseStyle() {
+  return css`
+    display: flex;
+    font-family: ${tokens.fontFamilies.sansSerif};
+    font-size: ${tokens.fontSizes.medium};
```
blurbyte, line 12: Don't think we need it anymore since List has font styles encapsulated and the previous content wrapper is no longer used (regular Paragraph is used instead).

src/empty-state/styles.ts
```diff
@@ -0,0 +1,106 @@
...
+    border-style: dashed;
+    border-radius: ${tokens.borderRadius.medium};
+    border-width: 1px;
```
blurbyte, line 15: Token can be used here `tokens.borderWidth.thin`.

src/empty-state/styles.ts
```diff
@@ -0,0 +1,106 @@
...
+  `;
+}
+
+export function listStyle() {
+  return css`
+    display: flex;
+    flex-direction: column;
+    margin: 0;
+    padding: ${tokens.spacing.medium} 0 ${tokens.spacing.large};
+    list-style: none;
```
blurbyte, line 88: If there is no custom icon provided for `ListItem` we should render default disc. Right now it's kind of hard to tell if it's a list:
Screenshot 2022-07-07 at 12 33 34

In this case padding left `tokens.spacing.large` looks good. Also `display: flex` will have to be disabled or disc won't render.

blurbyte (changes requested):
A few additional improvements could be made.

src/empty-state/styles.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type ListItemStyleOptions = {
+  hasIcon: boolean;
+};
+
+export function listItemStyle({ hasIcon }: ListItemStyleOptions) {
+  return css`
+    display: ${hasIcon ? 'flex' : 'list-item'};
+    list-style: ${hasIcon ? 'none' : 'inside'};
```
blurbyte, line 91: It looks pretty bad with `list-style: inside` if the content is long:
Screenshot 2022-07-08 at 16 34 50

I would change it to `outside` and add some `margin-left`. Either `tokens.spacing.large` if we want to keep nice indentation or `tokens.spacing.xlarge` if we want a text to be aligned for version with and without icon (not sure if they should/will be mixed at all).

doc-site/pages/components/empty-state.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+<Example title="Basic Usage" path="empty-state/basic-usage">
+  <BasicUsage />
+</Example>
+
+<Example title="Orientation" path="empty-state/orientation">
+  <Orientation />
+</Example>
+
+<Example title="Centered Column" path="empty-state/centered-column">
```
blurbyte, line 31: 'Centered' should be enough. Example file name could be changed to `centered.tsx` too.

doc-site/examples/empty-state/centered-column.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+function Example() {
+  return (
+    <EmptyState
+      title="You have no assignments yet"
+      image={<Paper3d />}
+      orientation="vertical"
+      isCentered
+    >
+      <Paragraph>
```
blurbyte, line 14: Maybe add some custom larger bottom margin here.

doc-site/examples/empty-state/inverted.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+function Example() {
+  return (
+    <EmptyState title="Title Heading" image={<Rocket3d />} inverted>
+      <Paragraph inverted>
+        Groups are most useful when you’re learning with your colleagues. Invite
+        members to your group here or share your unique invite link with them
+        directly.
+      </Paragraph>
+      <Paragraph>
```
blurbyte, line 15: This paragraph could be removed.

blurbyte (changes requested):
Just a few minor changes.

doc-site/pages/components/empty-state.mdx
ixTec: Going to rename it to `Vertical Centered` just to make it clear that this is the only use case for centered

src/empty-state/styles.ts
```diff
@@ -0,0 +1,104 @@
...
+
+type ListItemStyleOptions = {
+  hasIcon: boolean;
+};
+
+export function listItemStyle({ hasIcon }: ListItemStyleOptions) {
+  return css`
+    display: ${hasIcon ? 'flex' : 'list-item'};
+    list-style: ${hasIcon ? 'none' : 'outside'};
+    margin-left: ${hasIcon ? '0' : '16px'};
```
blurbyte, line 92: It still looks a bit weird. I would go with `margin-left: ${tokens.spacing.large}` or `margin-left: ${tokens.spacing.xlarge}` with `padding-left: 0;` when there is no icon. Spacing `large` looks better but spacing `xlarge` makes variants with/without icon text aligned (prolly we can ask Petra for feedback about it).

blurbyte (changes requested):
Looks really good, just 1 small change to `ListItem` and we are good to go.

blurbyte (approved):
Let's merge 🚀.

PR #56, opened 2022-06-27
[WF-618] Artefacts around border of component

ixTec:
# Fixes artefacts around border of component

### 🔧 Minor Changes

- Fixed artefacts around border / background of avatar wrapper


blurbyte (approved):
That's really smart.

I am wondering if we should use it for all avatars or only if `fillContent` is specified. It makes all avatars with solid fill 2px smaller (which is barely noticeable of course). The scenario when we will have 2 avatars: one regular and one with fill content next to each other is unlikely, but still.

So at the end to satisfy pixel-peepers I would enable the gradient for `fillContent` and leave regular one when it's not specified (and it's not an issue).

blurbyte (approved):
Looks nice, let's merge 🚀

PR #57, opened 2022-07-05
[WIP] Documentation Overhaul / Improvements

ixTec:
# Documentation Overhaul / Improvements

- [WF-608](https://datacamp.atlassian.net/browse/WF-608) ???
- [WF-609](https://datacamp.atlassian.net/browse/WF-609)
- [WF-610](https://datacamp.atlassian.net/browse/WF-610)
- [WF-611](https://datacamp.atlassian.net/browse/WF-611)
- [WF-613](https://datacamp.atlassian.net/browse/WF-613)

### 🚀 Major Changes

- Added new page header to top of page content
    - Category is determined from page url
    - Layout use within `.mdx` files has slightly changed
- Updated table of contents design
    - Shows active section, based on scroll position / elements in view of content

### 🔧 Minor Changes

- Added figma link to top navigation bar
- Added scroll to top link

### ❓ Misc Changes

- Original `page-header` has been renamed to `top-navigation`

## Screenshot(s):
Screenshot 2022-07-05 at 11 21 58
Screenshot 2022-07-05 at 11 24 02
Screenshot 2022-07-05 at 11 26 44


doc-site/components/table.tsx
```diff
@@ -18,10 +18,7 @@ const tableCellStyle = css`
   text-align: left;
   word-wrap: break-word;
   vertical-align: top;
-  padding-top: ${tokens.spacing.small};
```
blurbyte, line 21: I would revert these values at least for now. Regular table, e.g. Roadmap one looks a bit too heavy:
Screenshot 2022-07-07 at 14 29 13

doc-site/components/table-of-contents.tsx
```diff
@@ -2,37 +2,58 @@ import Link from 'next/link';
...
   name: string;
+  isActive: boolean;
 };
 
-function Entry({ name }: EntryProps) {
+function Entry({ name, isActive }: EntryProps) {
   return (
-    <li css={listItemStyle}>
+    <li css={listItemStyle({ isActive })}>
       <Link href={`#${slugify(name)}`} passHref>
```
blurbyte, line 57: With a border to the side and heading I think it will look a bit nicer if we use small `Button` as a link here.

doc-site/components/table-of-contents.tsx
```diff
@@ -41,17 +62,24 @@ function Entry({ name }: EntryProps) {
...
-  const entries = useTableOfContentsEntries();
+  const content = useTableOfContentsEntries();
 
-  if (entries.length > 0) {
+  if (content.entries.length > 0) {
     return (
-      <aside>
+      <div css={contentsTableStyle}>
+        <Heading as="h2" css={headingStyle}>
+          CONTENT
```
blurbyte, line 71: Could be capitalised via CSS.

doc-site/components/markdown-elements.tsx
```diff
@@ -26,37 +27,27 @@ type ContentProps = {
 function H1({ children }: ContentProps) {
```
blurbyte, line 27: Don't think we need it anymore, since H1 will be always provided with `PageHeader`.

doc-site/components/all-asset-types.tsx
```diff
@@ -14,16 +14,20 @@ const groupMaxColumnCount = {
...
-      key={groupName}
-      assetType={groupName}
-      assets={assetGroup}
-      maxColumns={
-        groupMaxColumnCount[groupName as keyof typeof groupMaxColumnCount]
-      }
-    />
-  ));
+  return (
+    <>
```
blurbyte, line 18: Does wrapping it with fragment fix the ordering?

blurbyte (changes requested):
Some general thoughts:

## Table of Contents (WF-609)

Deffo it's pretty close to being ready, although at the moment the active item selection seems a little buggy - sometimes section is not selected or skipped altogether. I think it would be best to extract this logic to separate story / branch and work on it later. For now I would keep styling and heading for sure.
On a side note this active item selection must be tricky to implement - check how janky it's here (it doesn't even work on first page load): https://atlassian.design/components/calendar/examples

## Figma Link (WF-610)

Looks good.

## Back to Top (WF-611)

I would show it a bit later, below a fold (could be approximate hardcoded value). You may add some nice fade in/out animation with `useAnimateTransition` hook too. Otherwise looks good.

## Props Table (WF-612)

Looks good. I would maybe consider showing multiple items in separate pills (not separated by pipes), like here:
Screenshot 2022-07-07 at 13 55 36

## Page Header (WF-613)

Looks really nice.
Would wrap it with `ErrorBoundary`. Maybe it would be possible to add the `pageTitle` and `pageDescription` to `PageLayout` actually if we have a header everywhere, and not passing `PageHeader` component as a prop on each page.
Also from what I remember for layout in MDX instead of wrapping whole page it's possible to do something like (not really sure if it's better than wrapping imho but something to consider for sure):
```
export default function Layout({children}) {
  return <PageLayout somePropsHere>{children}</PageLayout>;
}
```

I think we no longer need custom `Paragraph` styles in `markdown-elements.tsx`.

## Section Separator (WF-614)

Spacing is a bit off, should be more balanced as in Figma:
Screenshot 2022-07-07 at 14 23 45
I think it would be simpler to just add border above our custom H2 and call it a day. This way it won't be rendered below last item on the page:
Screenshot 2022-07-07 at 14 49 11

doc-site/components/markdown-elements.tsx
ixTec: Keeping it because we need to ensure we're setting the sizing / waffles component

doc-site/components/back-to-top.tsx
```diff
@@ -0,0 +1,42 @@
...
+  bottom: ${tokens.spacing.xlarge};
+`;
+
+function BackToTop() {
+  const [isVisible, setIsVisible] = useState(false);
+
+  useEffect(() => {
+    // Show button if the window is scrolled below the default header height (54px)
+    function handleVisibilityChange() {
+      window.scrollY > 54 ? setIsVisible(true) : setIsVisible(false);
```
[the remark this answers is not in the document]
ixTec: Keeping as is, as discussed, but replaced value with constant

doc-site/components/table.tsx
ixTec: I've overwritten the new styling for the component status and design token tables, but kept roadmap and other tables (props etc) as is for consistency for now. Can be reviewed again later

doc-site/components/back-to-top.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    handleVisibilityChange();
+    window.addEventListener('scroll', handleVisibilityChange);
+
+    return () => {
+      window.removeEventListener('scroll', handleVisibilityChange);
+    };
+  }, []);
+
+  return isVisible
```
blurbyte, line 41: `isAnimating` should be enough here.

blurbyte (changes requested):
Looks really good. Only a few things to correct left:

1. Component Status table looks off (left spacing is not aligned with other table)

Screenshot 2022-07-14 at 11 01 20

2. I would reduce spacing between the line and link in ToC even further (I think right now it's `medium` so let's make it `small`:

Screenshot 2022-07-14 at 11 03 03

3. Props look much better now. To make spacing between items more consistent maybe wrap them with a div with `display: flex; flex-wrap: wrap; gap: 2/4px whatever works better;`

Screenshot 2022-07-14 at 11 06 07

PR #59, opened 2022-07-13
WF-621 - Loader rendering issue in safari

ixTec:
# Fix loader rendering issue in safari

### 🔧 Minor Changes

- Fixed rendering issue of svg within loader in Safari browser

### ❓ Misc Changes

- Moved stroke property from path element to svg element, to solve issue with setting custom stroke value

## Screenshot(s):
Screenshot 2022-07-13 at 12 04 47


blurbyte (approved):
Nice. Checked on MacBook and iPhone and works as expected on both - let's merge 🚀.

PR #60, opened 2022-07-14
Small Menu and Dialog improvements

blurbyte:
# Destructive Menu.Item and Dialog.Footer improvements

### 🚀 Major Changes

- Added `destructive` variant to `Menu.Item`.

### 🔧 Minor Changes

- Updated the default alignment of buttons in `Dialog` and `Drawer` footers (so they are set apart instead close together).
- Increased the size of alert dot of `Menu.Item`.

### ❓ Misc Changes

- Reverted changes of custom Jest setup. Unfortunately it was breaking Cypress typings.
-  Bumped _emotion_ and _floting-ui_ deps versions.
- Updated main background color of docs based on updated specs.

## Screenshot(s):
Screenshot 2022-07-14 at 15 30 38
Screenshot 2022-07-14 at 15 31 08


doc-site/pages/overview/roadmap.mdx
```diff
@@ -6,14 +6,14 @@ import ComponentStatus from '../../components/component-status';
 
 ## Next Up
 
-_Last updated on 10 June 2022._
+_Last updated on 14 June 2022._
```
ixTec, line 9: Should be July

doc-site/examples/dialog/playground-config.ts
```diff
@@ -35,13 +35,21 @@ function Playground() {
...
             Cancel
           </Dialog.Button>
-          <Dialog.Button
-            autoFocus
-            variant="primary"
-            onClick={() => setIsOpen(false)}
-          >
-            Confirm
-          </Dialog.Button>
+          <div>
```
ixTec, line 38: On very small screen widths, the buttons in this div begin to wrap. We should add some styling in this playground to clean up how they look in this case.

On top of this, there is no wrapping behaviour for the footer, and so content is overflowing if it's too large. We should probably address this too
Screenshot 2022-07-14 at 15 13 50

src/menu/styles.ts
```diff
@@ -60,31 +62,56 @@ const itemBaseStyle = css`
...
+    color: tokens.colors.redDarkText,
+    hoverColor: hexToRgba(tokens.colors.redDarkText, tokens.opacity.low),
+    activeColor: tokens.colors.greySubtle,
+  },
+};
+
+const invertedVariantMap = {
+  primary: {
+    color: tokens.colors.white,
+    hoverColor: hexToRgba(tokens.colors.white, tokens.opacity.low),
```
ixTec, line 83: I would rename `hoverColor` / `activeColor `to better imply they're intended for backgroundColor use. I'm also unsure behind the naming of `activeColor` currently

ixTec (changes requested):
Few minor comments, with the primary one being the dialog button wrapping

doc-site/examples/dialog/playground-config.ts
blurbyte: Good catch.

ixTec (approved):
Let's mergeeee 🚀

PR #65, opened 2022-07-19
[WF-626] Asset docs heading order

ixTec:
# Fix ordering of headings in table of contents on asset docs page

### 🔧 Minor Changes

- Fixed order of headings on asset docs page

### ❓ Misc Changes

- Fixed margin issue on icon grid, by removing unnecessary section wrapper

## Screenshot(s):
Screenshot 2022-07-19 at 15 50 08


blurbyte (approved):
Awesome🥇! To me it looks even better than previously (markdown file is for sure more readable, with each asset group with each own section).

PR #66, opened 2022-07-22
[WF-603] Resizable

blurbyte:
# Resizable component

### 🚀 Major Changes

- Implemented `Resizable` component: it renders its children separated by draggable dividers, so the width or height (depending on orientation) could be adjusted:
  - orientation could be changed (vertical / horizontal)
  - initial proportions could be set (otherwise subsections are split equally)
  - minimum size of subsection could be set, and subsection can't be collapsed below this value
  - it's possible to show / hide separators depending on the situation
  - there a few workarounds in styles because linter complains when too many vars are used in template literals (e.g. `${dimension}: ${size}px` is not fine for linter)
  - ⚠️ ~how onResizeEnd is handled is not ideal (it's called in useEffect which could cause memory leak); find another way to expose current proportions~

**Questions**
- What is better name: _divider_ or _separator_ (`showSeparators` prop)
- Is `minSubsectionSize` clear enough? Maybe there is better name for _subsection_?

## Screenshot(s):

Nested layout example:
resizable-example

Recording:
https://user-images.githubusercontent.com/20565536/180423674-1872edf6-4e6f-4b66-976b-dbdefc8e09ad.mp4

### DIFF
--- a/doc-site/examples/resizable/playground-config.tsx
+++ b/doc-site/examples/resizable/playground-config.tsx
@@ -18,7 +18,7 @@ function Playground() {
         backgroundColor: tokens.colors.beigeLight,
       }}
     >
-      <Resizable initialProportions={[60, 40]} showDividers>
+      <Resizable defaultProportions={[60, 40]} showDividers>
         <Paragraph style={{ padding: tokens.spacing.medium }}>
           First Panel
         </Paragraph>
--- a/doc-site/examples/resizable/playground-config.tsx
+++ b/doc-site/examples/resizable/playground-config.tsx
@@ -5,6 +5,7 @@ import { Paragraph } from '@datacamp/waffles/paragraph';
 import type { PlaygroundConfig } from '../../types';
 
 const initialCode = `
+import { tokens } from '@datacamp/waffles/tokens';
 import { Resizable } from '@datacamp/waffles/resizable';
 import { Paragraph } from '@datacamp/waffles/paragraph';
 
--- a/doc-site/examples/resizable/playground-config.tsx
+++ b/doc-site/examples/resizable/playground-config.tsx
@@ -1,0 +1,43 @@
+import { Resizable } from '@datacamp/waffles/resizable';
+import { Paragraph } from '@datacamp/waffles/paragraph';
+
+import type { PlaygroundConfig } from '../../types';
+
+const initialCode = `
+import { Resizable } from '@datacamp/waffles/resizable';
+import { Paragraph } from '@datacamp/waffles/paragraph';
+
+function Playground() {
+  return (
+    <div
+      style={{
+        width: '100%',
+        height: '200px',
+        borderRadius: '4px',
+        backgroundColor: '#F7F3EB',
+      }}
+    >
+      <Resizable initialProportions={[60, 40]} showSeparators>
+        <Paragraph style={{ padding: '16px' }}>
+          First Resizable Subsection
+        </Paragraph>
+        <Paragraph style={{ padding: '16px' }}>
+          Second Resizable Subsection
+        </Paragraph>
+      </Resizable>
+    </div>
+  );
+}
+
+export default Playground;
+`;
+
+const playgroundConfig: PlaygroundConfig = {
+  initialCode,
+  scope: {
+    Resizable,
+    Paragraph,
+  },
+};
+
+export default playgroundConfig;
--- a/doc-site/examples/resizable/playground-config.tsx
+++ b/doc-site/examples/resizable/playground-config.tsx
@@ -1,3 +1,4 @@
+import { tokens } from '@datacamp/waffles/tokens';
 import { Resizable } from '@datacamp/waffles/resizable';
 import { Paragraph } from '@datacamp/waffles/paragraph';
 
@@ -13,16 +14,16 @@ function Playground() {
       style={{
         width: '100%',
         height: '200px',
-        borderRadius: '4px',
-        backgroundColor: '#F7F3EB',
+        borderRadius: tokens.borderRadius.medium,
+        backgroundColor: tokens.colors.beigeLight,
       }}
     >
-      <Resizable initialProportions={[60, 40]} showSeparators>
-        <Paragraph style={{ padding: '16px' }}>
-          First Resizable Subsection
+      <Resizable initialProportions={[60, 40]} showDividers>
+        <Paragraph style={{ padding: tokens.spacing.medium }}>
+          First Panel
         </Paragraph>
-        <Paragraph style={{ padding: '16px' }}>
-          Second Resizable Subsection
+        <Paragraph style={{ padding: tokens.spacing.medium }}>
+          Second Panel
         </Paragraph>
       </Resizable>
     </div>
@@ -37,6 +38,7 @@ const playgroundConfig: PlaygroundConfig = {
   scope: {
     Resizable,
     Paragraph,
+    tokens,
   },
 };
 
--- a/src/resizable/constants.ts
+++ b/src/resizable/constants.ts
@@ -1,2 +1,3 @@
-export const DIVIDER_SIZE = 6;
+export const DIVIDER_HITBOX_SIZE = 7;
+export const DIVIDER_LINE_SIZE = 1;
 export const KEYBOARD_STEP = 10;
--- a/src/resizable/constants.ts
+++ b/src/resizable/constants.ts
@@ -1,2 +1,2 @@
-export const DIVIDER_WIDTH = 6;
+export const DIVIDER_SIZE = 6;
 export const KEYBOARD_STEP = 10;
--- a/src/resizable/constants.ts
+++ b/src/resizable/constants.ts
@@ -1,1 +1,2 @@
 export const DIVIDER_WIDTH = 6;
+export const KEYBOARD_STEP = 10;
--- a/src/resizable/constants.ts
+++ b/src/resizable/constants.ts
@@ -1,0 +1,1 @@
+export const DIVIDER_WIDTH = 6;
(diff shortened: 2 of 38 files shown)


src/resizable/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const DIVIDER_HITBOX_SIZE = 7;
+export const DIVIDER_SEPARATOR_SIZE = 1;
+export const KEYBOARD_STEP = 10;
```
ixTec, line 3: I think we may want these configurable by the consumer?
Especially  `DIVIDER_HITBOX_SIZE` or `DIVIDER_HITBOX_SIZE`

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,319 @@
...
+    dimension: 'height',
+    offset: 'y',
+    mousePosition: 'clientY',
+    keyPositive: 'ArrowDown',
+    keyNegative: 'ArrowUp',
+  },
+} as const;
+
+type ResizableProps = {
+  /* The elements to render in subsections, divided by separator. At lest 2 must be provided. */
```
ixTec, line 40: "... Must provide at least 2 (two) elements."

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,319 @@
...
+  children: JSX.Element[];
+  /* The layout of the subsections. */
+  orientation?: 'vertical' | 'horizontal';
+  /* An array of proportions, e.g. `[2, 1, 1]`, which determine initial size of each subsection. Must have the same length as the number of provided elements. When not provided the subsections will default to equal sizes. */
+  initialProportions?: number[];
+  /* The minimal size of the subsection. Default is `100px`. */
```
ixTec, line 46: Change to "The minimum size..."

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,319 @@
...
+    keyNegative: 'ArrowUp',
+  },
+} as const;
+
+type ResizableProps = {
+  /* The elements to render in subsections, divided by separator. At lest 2 must be provided. */
+  children: JSX.Element[];
+  /* The layout of the subsections. */
+  orientation?: 'vertical' | 'horizontal';
+  /* An array of proportions, e.g. `[2, 1, 1]`, which determine initial size of each subsection. Must have the same length as the number of provided elements. When not provided the subsections will default to equal sizes. */
```
ixTec, line 44: "... which determine the default relative size of each section. ..."

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,319 @@
...
+  initialProportions?: number[];
+  /* The minimal size of the subsection. Default is `100px`. */
+  minSubsectionSize?: string;
+  /* If enabled, separators between subsections are visible. */
+  showSeparators?: boolean;
+  /* Sets the style of the separator suitable for dark backgrounds. */
+  inverted?: boolean;
+  /* Handler called when separator just started to move. */
```
ixTec, line 52: Can change comment to:
"Called when divider begins to move."

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,319 @@
...
+  onResizeStart?: () => void;
+  /* Handler called when separator just stopped moving. Useful for retreiving subsections proportions after resizing via `proportions` argument. For example could be used to persist proportions in local storage. */
```
ixTec, line 54: Think we can remove the "for example..." bit and update this comment to be:

"Called when divider stops moving. Useful for retrieving the proportions of sections." I don't think it's currently clear what the `proportions` argument is here

src/resizable/subsection.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+function Subsection({
+  children,
+  orientation,
+  isDragging,
+  compensateForSeparator,
+  dimension,
+}: SubsectionProps) {
+  // Because of frequent updates, pass width / height as regular style
+  // It prevents emotion from appling new class each time a new value is provided
```
ixTec, line 24: "applying"

ixTec (changes requested):
Some very creative solutions to tricky problems here, nice!

Added several comments to the code, but also some overall points too:

**Answers to questions:**
- I'm leaning towards `divider` as a term, so `showDividers`. etc
- In regards to `minSubsectionSize`, I think the documentation should mention it's width vs height depending on orientation. I think also something can only be a "subsection" if it's a child of another subsection, so I think "section" is enough. Likely `minSize` is enough here too though.

**Misc comments/feedback:**
- In regards to terminology, I am not keen on the "subsection"/"section" use. I think one of the following may be more suitable: "segment" or "panel" - what do you think?
- Would we benefit from a 'maxSize' prop also?
- I would rename `initialProportions` to `defaultSizes`
- There is quite a large issue when resizing the browser window (and thus changing the size of the resizable container). It does not update measurements of sections correctly. Happy to demonstrate this if needed
- I think there's some room for confusion over the `orientation` and how it impacts the layout. "Vertical" `orientation` splits the component into columns (and so across the container), but horizontal `orientation` splits it into rows, down the container ? Perhaps we need to reconsider this, especially if we compare to something like empty-state
- Looks like some of the types in the props table aren't resolving correctly (e.g. `children`)
- The labels used within the doc examples seem a bit mixed up. Some "First Second Subsection" and lots of "First Resizable Subsection" duplication in examples. Also, some should probably be "Section" vs "Subsection"

I still more comments to come next week on the actual component implementation

blurbyte:
Great review, a lot of good points. Based on feedback I'm planning to:

- Rename Separator -> Divider
- Reame Subsection -> Panel
- I think `initialProportions` better reflect, what it actually does; I feel `defaultSizes` kind of implies you want to provide sizes in pixels
- Good catch about not handling browser window resize - in this case proportions should be preserved
- About `orientation`: `row` / `column` is a possibility here for sure
- Children are `JSX.Element[]` so its technically `arrayType`

Something for future releases:

- `maxSize`

src/resizable/constants.ts
blurbyte: Yeah, something to consider for future releases, maybe even providing custom divider component. But overall it should be rather rare use case. In general we want all resizables to look the same.

doc-site/examples/resizable/playground-config.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const initialCode = `
+import { Resizable } from '@datacamp/waffles/resizable';
+import { Paragraph } from '@datacamp/waffles/paragraph';
+
+function Playground() {
+  return (
+    <div
+      style={{
+        width: '100%',
```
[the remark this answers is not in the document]
blurbyte: That's what I tried at first, but turns out emotion was going crazy and applied new class each time value changed when dragging, which you can imagine affected performance by a lot (made it laggy).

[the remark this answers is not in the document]
blurbyte: For some reason emotion is not working in playground so have to use regular `style`.

doc-site/examples/resizable/playground-config.tsx
```diff
@@ -0,0 +1,43 @@
...
+        height: '200px',
+        borderRadius: '4px',
```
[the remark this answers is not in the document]
blurbyte: Will check it, not 100% sure if will work in playground.

ixTec (changes requested):
Looking much better!

- Uses of `subsection` should be updated to be `panel`
- Uses of `separator` should be updated to be `divider`
- There's some funky behaviour around resizing after resizing another one. It's almost like it allows you resize beyond the original min size? (see attached video
- Previously discussed resize behaviour

https://user-images.githubusercontent.com/9938773/180987499-5b9b1f14-e2f8-4c9f-afc4-d8cb81cf12d1.mov

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+        setPanelsDimensions((previousDimensions) => {
+          if (containerRef.current && currentDividerIndex.current !== null) {
+            const updatedDimensions = previousDimensions;
+            const dividerIndex = currentDividerIndex.current;
+            const containerBoundingBox =
+              containerRef.current.getBoundingClientRect();
+
+            // Set min and max divider positions, so it can't be dragged beyond container boundaries
+            let minDividerPosition = 0;
```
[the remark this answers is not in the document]
blurbyte: Really good suggestion, makes it much more readable.

doc-site/examples/resizable/playground-config.tsx
```diff
@@ -0,0 +1,45 @@
+import { tokens } from '@datacamp/waffles/tokens';
+import { Resizable } from '@datacamp/waffles/resizable';
+import { Paragraph } from '@datacamp/waffles/paragraph';
+
+import type { PlaygroundConfig } from '../../types';
+
+const initialCode = `
+import { Resizable } from '@datacamp/waffles/resizable';
+import { Paragraph } from '@datacamp/waffles/paragraph';
```
ixTec, line 9: tokens should be included as an import in the initial code for the example

src/resizable/resizable.tsx
```diff
@@ -0,0 +1,367 @@
...
+} as const;
+
+type ResizableProps = {
+  /* The elements to render in separate panels, divided by separator. Must provide at least _two_ elements. */
+  children: JSX.Element[];
+  /* The layout of the panels. */
+  layout?: 'column' | 'row';
+  /* An array of proportions, e.g. `[2, 1, 1]`, which determine the default relative size of each panel. Must have the same length as the number of provided elements. When not provided the panels will default to equal sizes. */
+  defaultProportions?: number[];
+  /* The minimum size of the panel. When resizing, panel can't be collapsed below this value. Default is `100px`. */
```
ixTec, line 48: Should mention that this value should be specified only in pixels

ixTec (changes requested):
I really like the advanced usage example! Nice one.

Approved with some minor changes required before merge (some could be put into the refinement ticket):

- Missing import inside playground example
- `minSize` prop description should mention that this value should be specified in pixels
- If a `minSize` value is provided that is larger than the width of the component, it does not look good :D

PR #69, opened 2022-07-28
[WF-551] Add `isLoading` state to button

ixTec:
# Add `isLoading` state to button

### 🚀 Major Changes

- Add new `isLoading` prop for button component
    - Button will maintain width from original content, unless `loadingLabel` is provided and as larger than original content
- Added `loadingLabel` prop
     - Optional prop to provide a label to the button in the loading state

### ❓ Misc Changes

- Fix a couple of minor ts issues
- Fixed an issue with the rendering of the `Loader` component

## Screenshot(s):
Screenshot 2022-07-28 at 14 52 47

https://user-images.githubusercontent.com/9938773/181522787-546b5d8f-6f41-4067-8072-1a02d11ba01e.mov


doc-site/pages/components/button.mdx
```diff
@@ -134,6 +135,10 @@ For standalone navigational actions, it is fine to use button&mdash;just set com
   <AsALink />
 </Example>
 
+<Example title="Is Loading" path="button/is-loading">
```
blurbyte, line 138: I think it shouldn't be standalone example but part of 'States'.

src/button/__tests__/button.spec.tsx
```diff
@@ -202,6 +202,19 @@ describe('Button', () => {
     expect(icons).toHaveLength(2);
   });
 
+  it('renders loader with isloading prop', () => {
```
blurbyte, line 205: Should be `isLoading`

src/button/button-internal.tsx
```diff
@@ -73,6 +88,7 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
   return (
     <Element
       {...mergeProps(focusProps, restProps)}
+      {...(isLoading ? { disabled: true } : null)}
```
blurbyte, line 91: Maybe `{...(isLoading && { disabled: true }}`?

blurbyte (changes requested):

blurbyte (changes requested):
Looks nice!

A few things to update:
1. When it's icon-only button (some icon provided via `icon` prop) it should be square. if  I think removing paddings on sides and setting width to proper `sizing` depending on button size should do the trick.
2. Tbh I think 'Cannot be provided alongside the iconLeft property' is unnecessarily restrictive - just replace the content of the button when it's loading whether icon was provided or not. Provided icons shouldn't be displayed when loading. It's similar for Input, when you specify `type=search` it 'replaces' leftIcon with looking glass.
3. Something we might do in future, not necessarily now: I think it might be worth 'saving' the width of the button before loading so it won't change the size when loading is triggered. It's pretty annoying when buttons change their sizes depending on state.

lensbart:
Two minor typographic remarks:
* No space between ellipsis and preceding word
* Use the proper character `…` instead of three dots `...`

src/button/button-internal.tsx
```diff
@@ -73,21 +86,23 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
   return (
     <Element
       {...mergeProps(focusProps, restProps)}
+      {...(isLoading && { disabled: true })}
```
blurbyte, line 89: We will need to handle `aria-label` here to so the 'Loading' text would be appended to original one, while loading.

src/button/button-internal.tsx
```diff
@@ -73,21 +86,23 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
+        isLoading,
         isFocusVisible,
       })}
     >
       {icon ? (
         renderIcon(icon)
       ) : (
         <>
-          {iconLeft && renderIcon(iconLeft)}
+          {iconLeft && !isLoading && renderIcon(iconLeft)}
```
blurbyte, line 105: It could be good idea to create `renderLeftIcon` helper and just based on scenario render either icon / loader / nothing in this slot.

src/button/button-internal.tsx
```diff
@@ -102,6 +117,17 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
           {iconRight && renderIcon(iconRight)}
```
blurbyte, line 117: Could be rendered conditionally `!isLoading && iconRight && renderIcon(iconRight)`, it will make it clearer what is happening rather than applying CSS style which is coming from parent.

src/button/button-internal.tsx
```diff
@@ -73,21 +86,23 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
     >
       {icon ? (
         renderIcon(icon)
       ) : (
         <>
-          {iconLeft && renderIcon(iconLeft)}
+          {iconLeft && !isLoading && renderIcon(iconLeft)}
           {children && (
             <span
               css={innerContentStyle({
```
blurbyte, line 108: Here I would add `isLoading` option and apply those height / width styles based on it.

src/button/button-internal.tsx
```diff
@@ -102,6 +117,17 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
       )}
+      {isLoading && (
+        <div
+          css={loaderWrapperStyle({ size, hasLoadingLabel: !!loadingLabel })}
+        >
+          <Loader
+            css={loaderStyle({ variant, inverted })}
+            width={size === 'small' ? '12' : '16'} // Setting width prop here so that it gets passed to the loader as restProps for the svg element
+          />
+          {loadingLabel && <span>{loadingLabel}</span>}
```
blurbyte, line 128: For now (unless someones complain) I would just hardcode 'Loading...' label. Could use exact the same styling as for `innerContentStyle`, or just add padding to the left to compensate for loader (which is technically `iconLeft`) which will always be there.

src/loader/keyframes.ts
```diff
@@ -8,18 +8,17 @@ export const stroke = keyframes`
 
 export const loaderStartMask = keyframes`
   0%, 6% {
-    clip-path: polygon(0% -12%, 0% -12%, 169% 65.5%, 169% 65.5%);
+    clip-path: polygon(0% -5%, 0% 178%, 188% 106%, 178% 69%);
```
blurbyte, line 11: Changing the first percentage value from 0% to -10% for all those mask styles fixed for me this weird 'cut edge' issue:

Before
Screenshot 2022-08-03 at 13 52 58
After
Screenshot 2022-08-03 at 13 52 44

blurbyte (changes requested):
I think that, at least for now, we don't really need `loadingLabel` - it might lead to unnecessary inconsistencies. Just hardcoded 'Loading...' string should be enough for now. Not really sure why someone would want it customized.

It's kind of good for accessibility to always show this loading label. Screen reader announces it properly as well.
For icon-only button `aria-label` should be updated to `aria-label={`${previous-aria-label} Loading`}`.

I think it might be good idea to make handling of loading a bit more clearer internally, so instead of hiding everything with CSS applied to parent, just render it conditionally and apply styles explicitly.

Also investigated a bit issue with loader SVG (it was driving me crazy 😜 ) - it seems adding some negative percentages to polygon mask in `loaderStartMask`, and `loaderEndMask` fixed it for me. Not sure if it won't break anything else.

Not entirely sure if it's me but animation seems super slow. For regular `Loader` it takes 5 seconds to complete and here it's like 10 or so.

src/button/button-internal.tsx
ixTec: We can't conditionally render any of the original content if we want the width to be maintained when in loading state. We instead have to hide it via opacity

src/button/button-internal.tsx
ixTec: Regarding loadingLabel, Petra wanted it customisable which is why I added it as an optional prop. I think we have an issue by showing a default label because if the original content is short then the default loading label would make the button width grow - something we want to avoid right?

src/button/button-internal.tsx
ixTec: Not sure I understand this? We don't want to modify the original content at all when in loading state

src/button/button-internal.tsx
ixTec: We still want an iconLeft to render even in loading view (as it takes up width within the button, but will be hidden via opacity)

src/button/button-internal.tsx
blurbyte: Yes, you're right. It will make it shorter if we will remove it from DOM.

blurbyte (changes requested):
There are a few issues left to fix:

- my suggestion with conditional rendering was wrong, since components will be removed from DOM and width won't be preserved, so I guess the easiest solution would be, while loading, to wrap whole content (icons, label) with some `span` and set height and opacity to 0. So ideally button width when loading should match the width when not loading. At lest for now we don't care about case when content is shorter.
Screenshot 2022-08-04 at 14 42 22

- there is little problem with loader, it's now slightly cut diagonally at the top

src/button/button-internal.tsx
```diff
@@ -57,8 +70,8 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
   ref?: PolymorphicRef<T>,
 ) {
   const Element = as || 'button';
-
   const { focusProps, isFocusVisible } = useFocusRing();
+  const { 'aria-label': ariaLabel } = restProps;
```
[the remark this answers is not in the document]
ixTec: Not sure I see what the difference in this vs my current approach is?

src/loader/__tests__/__snapshots__/loader.spec.tsx.snap
```diff
@@ -67,7 +68,7 @@ exports[`Loader renders a loader 1`] = `
   >
     <svg
       class="emotion-2"
-      viewBox="0 0 2640 3444"
+      viewBox="0 0 2569 3403"
```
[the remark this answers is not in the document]
ixTec: Viewbox appeared to be incorrect based on the svg paths, so I had to fix this

blurbyte:
In general we stick to either `default` or `pointer`. Other cursors are fun for sure, but feel rather cheap, and so 2000. Not sure we need more cues here - there is loader and huge loading label. I can ask designers about their opinion next week.

src/menu/button.tsx
```diff
@@ -40,13 +40,12 @@ function MenuButton<T extends React.ElementType = 'button'>({
...
 
   return (
-    // @ts-expect-error: spread props out of sync
     <Button
       {...getItemProps({
         onClick: handleClick,
         ...restProps,
       })}
-      ref={(node) => (listRef.current[index] = node)}
+      ref={(node: HTMLButtonElement) => (listRef.current[index] = node)}
```
blurbyte, line 48: Good catch 👍

src/button/button-internal.tsx
```diff
@@ -102,6 +119,15 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
           {iconRight && renderIcon(iconRight)}
         </>
       )}
+      {isLoading && (
+        <div css={loaderWrapperStyle({ size, hasLoadingLabel: !icon })}>
```
blurbyte, line 123: I think it may be good idea to create separate internal component out of it.

src/button/button-internal.tsx
```diff
@@ -11,6 +18,10 @@ type ButtonBaseProps = {
   variant?: 'primary' | 'secondary' | 'plain' | 'destructive' | 'upgrade';
   /* Defines the size of the button. In most cases default `medium` size should be used. */
   size?: 'small' | 'medium' | 'large';
+  /* Whether the button is in it's loading state. The button will maintain the width of it's child content, unless `loadingLabel` is provided and longer than the original content. */
+  isLoading?: boolean;
+  /* An optional label to be shown whilst `isLoading` is set to true. */
+  loadingLabel?: React.ReactNode;
```
blurbyte, line 24: We don't need it anymore.

blurbyte (changes requested):
We are really close, there is still one issue:

Screenshot 2022-08-05 at 15 34 15

Because when loading `flex-direction` is changed elements are stacked vertically under rather next to each other, the width is not preserved. I think we have 2 possible solutions here: either add wrapper for before-loading content or position loader content absolutely in the center of the button.

Besides that everything looks good.

src/button/button-internal.tsx
```diff
@@ -73,21 +85,26 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
+        isLoading,
         isFocusVisible,
       })}
     >
       {icon ? (
         renderIcon(icon)
       ) : (
         <>
-          {iconLeft && renderIcon(iconLeft)}
+          {iconLeft && !isLoading && renderIcon(iconLeft)}
```
blurbyte, line 107: I was wrong with this one. In current setup, we shouldn't render it conditionally. `iconLeft && !isLoading && renderIcon(iconLeft)` is fine here.

blurbyte (changes requested):

src/button/loader.tsx
```diff
@@ -0,0 +1,27 @@
...
+type ButtonLoaderProps = {
+  size: NonNullable<React.ComponentProps<typeof Button>['size']>;
+  variant: NonNullable<React.ComponentProps<typeof Button>['variant']>;
+  inverted: NonNullable<React.ComponentProps<typeof Button>['inverted']>;
+  hasIcon: boolean;
+};
+
+function ButtonLoader({ size, variant, inverted, hasIcon }: ButtonLoaderProps) {
+  return (
+    <div css={loaderWrapperStyle({ size, hasLoadingLabel: !hasIcon })}>
```
blurbyte, line 17: Ideally we should use `<span>` here since technically it's more semantically correct to use it within in button element.

src/button/loader.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Loader } from '../loader';
+
+import { loaderStyle, loaderWrapperStyle } from './styles';
+import Button from './button';
+
+type ButtonLoaderProps = {
+  size: NonNullable<React.ComponentProps<typeof Button>['size']>;
+  variant: NonNullable<React.ComponentProps<typeof Button>['variant']>;
+  inverted: NonNullable<React.ComponentProps<typeof Button>['inverted']>;
+  hasIcon: boolean;
```
blurbyte, line 12: I would change it to `isIconOnly`. I think it better describes what's the purpose of it.

src/button/button-internal.tsx
```diff
@@ -73,34 +77,36 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
         hasIcon: !!icon,
+        isLoading,
         isFocusVisible,
       })}
     >
       {icon ? (
         renderIcon(icon)
       ) : (
-        <>
+        <span
```
blurbyte, line 98: So much simpler, thanks to `gap` 👏 .

src/button/button-internal.tsx
```diff
@@ -73,34 +77,36 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
+          css={innerContentStyle({
+            size,
```
blurbyte, line 100: I would add `isLoading` here instead of using `& > *:not(div)` in styles.

src/button/button-internal.tsx
```diff
@@ -3,6 +3,7 @@ import { mergeProps } from '@react-aria/utils';
 import { useFocusRing } from '@react-aria/focus';
 
 import { buttonStyle, innerContentStyle } from './styles';
+import ButtonLoader from './loader';
```
blurbyte, line 6: Probably can call it just `Loader` since there is no clash with our asset.

src/button/styles.ts
```diff
@@ -165,6 +167,7 @@ export function buttonStyle({
     width: ${fullWidth ? '100%' : 'auto'};
     padding-left: ${!hasIcon && sizeMap[size].spacing};
     padding-right: ${!hasIcon && sizeMap[size].spacing};
+    flex-direction: ${isLoading ? 'column' : 'row'};
```
blurbyte, line 170: Is it still even needed? I guess `column` should handle both.

blurbyte (changes requested):
Looks great, just a few minor tweaks. Really like usage of `gap` in content wrapper. Makes things much cleaner.

src/button/button-internal.tsx
```diff
@@ -73,34 +77,36 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
-              {children}
-            </span>
-          )}
+          {children}
           {iconRight && renderIcon(iconRight)}
-        </>
+        </span>
+      )}
+      {isLoading && (
+        <ButtonLoader size={size} variant={variant} inverted hasIcon={!!icon} />
```
blurbyte, line 109: When there are a lot of props with same names we can use `{...{ size, variant, inverted, isIconOnly: !!icon }}`.

blurbyte (changes requested):

src/button/button-internal.tsx
ixTec: Keeping it as `ButtonLoader` to save us having to use an alias within the button loader component. I don't think it makes any difference in this case and helps avoid us accidentally importing it when using loader internally 😄

src/button/button-internal.tsx
ixTec: Good idea, cheers!

src/button/button-internal.tsx
ixTec: Leaving as is, from our discussion

src/button/styles.ts
ixTec: Good catch, `column` should now work for both as we've wrapped the content

blurbyte (approved):
Looks great! Let's merge 🚀.

PR #71, opened 2022-07-29
Make `Tabs.Tab` children optional

lensbart:
(No ticket)

When `activeTab` is determined by the URL’s pathname rather than the hash, the concept of having `children` in `Tabs.Tab` doesn’t make sense, as the entire page will change upon navigation. In this situation, `children` should be optional. I think it would be too cumbersome to tell this case apart from other cases where `children` are in fact required (and in a sense, having an empty tab should be possible, too), so I didn’t attempt to.

**🔧 Minor Changes**

* Make the `children` prop optional

**To do**

* ~~Update the example in the Waffles docs to remove `children` for the “As Links” example, or provide an additional example where the entire page is the state~~ done
* ~~Remove the “required indicator” (red dot) at https://waffles-next.datacamp.com/components/tabs/#props~~ edit: not necessary


blurbyte:
@lensbart You don't have to do anything to remove required dot. Props table is generated automatically.

lensbart:
Even better! Thanks for the quick response. This is not urgent, make sure to enjoy your holiday! 😎

lensbart:
@ixTec
Thanks, Christian. I saw you added me as an assignee. Was this by mistake? (I don’t think I’m supposed to review my own PR?)

I looked through the contributing guidelines:
> Below are a handful of simple rules [...] for those that ESLint / Prettier won't handle.

Some of the coding style guidelines can be enforced by ESLint. Make sure to check out the following links.
* Functions
  * https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md
  * Perhaps also https://eslint.org/docs/latest/rules/func-names & https://eslint.org/docs/latest/rules/func-style#allowarrowfunctions but I think these are not sufficient or less relevant
  * https://eslint.org/docs/latest/rules/prefer-arrow-callback#rule-details
* [Enforcing file names](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md)
* TODO format: feel free to take some inspiration from [our custom rule](https://github.com/datacamp-engineering/collab-and-tooling/blob/master/libs/config/eslint-plugin-collab/src/rules/smartTodo.ts)

As an aside, I saw that a comment of mine on [a different PR](https://github.com/datacamp/waffles/pull/69) was deleted. Was this to indicate that it was resolved? I had to look through the code to see what came of it (apparently there will be a `loadingLabel` prop).

ixTec:
Added you as the assignee to follow up with finishing the TODO in your PR description and to ensure it's following our contribution guidelines (mainly, branching and commit messages etc) _before_ we review it. Feel free to mark the PR as a draft whilst this happens if you'd prefer.

You're right about some of the ESLint configurations, but we've not yet added these to our config, which is why they're included in the current guidelines. 

As for the deleted comment, it was to indicate that it was no longer applicable for the approach we were taking with the component. I was a bit surprised to see a comment on a PR from outside the team, but appreciated the heads up regarding the ellipsis character so cheers 👍

lensbart:
I meant the TODO more as something that might be done separately. I don’t think it’s needed for this PR to land, as the examples remain valid. Is it okay if I file a ticket in Jira?

I updated the base branch just now.

ixTec:
Happy for a ticket to be raised for the example update. Could you also update the commit name/message to include the component name too please? (`fix(button): make children optional`). 

Otherwise looks good to me

ixTec (approved):
Looks good, let's merge

lensbart:
Whoops, in the meantime I took a quick stab at editing the docs. You may want to give it a quick look before it gets merged.

doc-site/pages/components/tabs.mdx
```diff
@@ -34,6 +34,13 @@ import Inverted from '../../examples/tabs/inverted';
   <AsLinks />
 </Example>
 
+<Example
+  title="As Links With No Children"
+  path="tabs/as-links-with-no-children"
+>
+  <AsLinksToDifferentPages />
```
ixTec, line 41: You'll have to import the example file at the top too, presumably this should be `<AsLinksWithNoChildren />` and
`import AsLinksWithNoChildren from '../../examples/tabs/as-links-with-no-children';`

doc-site/examples/tabs/as-links-with-no-children.tsx
```diff
@@ -0,0 +1,46 @@
+import { useState, useEffect } from 'react';
```
ixTec, line 1: In reflection, it may be better suited to just provide a "no-tab-content" example (or similar), instead of another links example. I'd probably just have something along the lines of the following:
```jsx
const [activeTabIndex, setActiveTabIndex] = useState<React.Key>(0);

  return (
    <>
      <Tabs
        activeTab={activeTabIndex}
        onChange={(activeTab) => {
          setActiveTabIndex(activeTab);
        }}>
        <Tabs.Tab label="First" />
        <Tabs.Tab label="Second" />
      </Tabs>
      <Paragraph>
        For use when the tabs have no content but instead control the context of the whole page.
      </Paragraph>
    </>
  );
  ```
  
This is also agnostic to how the individual tabs are used (whether as links or not). I would include the brief paragraph below the example (similar to our Advanced Usage example for our new [Resizable](https://waffles-next.datacamp.com/components/resizable/) component.

ixTec (changes requested):

blurbyte:
@lensbart Thanks for advice on linting rules, we will add them rather soon.

ixTec (approved):
Approved, good to merge

PR #74, opened 2022-08-04
[WF-643] Fix console warning for menu component

lensbart:
Fixes an issue where a boolean value (`inverted`) was inadvertently passed to an HTML element.

The type of `getReferenceProps` is `(userProps?: React.HTMLProps<Element> | undefined) => Record<string, unknown>`, so its argument should be an object with only valid HTML attributes as key/value pairs.

### ❓ Misc Changes

- Fixes a console warning


blurbyte:
Thanks for reporting it. Unfortunately removing injected `inverted` attribute is not ideal here, since it breaks dark mode. I have simple fix in mind tho (will investigate it at the beginning of the next week).

blurbyte:
@lensbart @ixTec Updated it slightly, so if the flag is not specified it won't be passed at all, and that warning won't appear:
Screenshot 2022-08-08 at 10 32 01
Previously it was always passed to trigger component whether it was set or not.
For _inverted_ variant (which we are not using a lot) we assume trigger handles it somehow - I think it is good compromise.

blurbyte (approved):

PR #75, opened 2022-08-05
[WF-608] Simplify docs layout

blurbyte:
# Simplify docs layout with grid

### 🔧 Minor Changes

- Removed redundant `overflow: hidden` from content container, so position `fixed` and `sticky` will work properly. Doesn't look like it broke anything.
- Simplified and streamlined docs layout with grid (compared to mockups changed `gap` to be slightly larger, aside was too close to content otherwise). It looks and behaves very similar to previous implementation.
- Lay out `PageHeader` in similar way to main content


ixTec (changes requested):
Looks good to me, grid layout is much simpler in this case.

I think we need to update the width/layout of the page header now too, to match the column width of the content

blurbyte:
You are right! Completely forgot about it.

ixTec (approved):
Happy to merge 👍

PR #77, opened 2022-08-09
[WF-550] `Progress` component

ixTec:
# Add new `Progress` component

### 🚀 Major Changes

- Added new `Progress` component

## Questions/Issues:

1. I've added `max` and `value` to the documented props, even though they're the props on a native progress bar. We can leave it as is, to ensure we're clear about the component API, or remove and then use our generic _"`Progress` accepts all props regular `HTMLProgress` would, such as: `max` and `value`. "_. - What do you think?
2. ~~**TODO:** Component is currently not responsive in Chrome, possibly solved by using a `mask` instead of `clip-path`, but haven't tried this yet. Looking for other solutions also.~~
3. Do we want an option to show/hide the label (with an aria-label provided instead)? Probably a stretch/future task though.

## Screenshot(s):
Screenshot 2022-08-09 at 09 47 55
Screenshot 2022-08-09 at 09 48 35
Screenshot 2022-08-09 at 09 48 47


blurbyte:
To answer the questions:

1. I think it is good idea, progress in not really commonly used element and ppl are not familiar with it's native API.
2. Checking it too.
3. Seems like pretty useful feature to have from the start.

src/progress/progress.tsx
```diff
@@ -0,0 +1,63 @@
...
+  /* The current value of the progress. This is in relation to the `max` property. */
+  value?: number;
+  /* The maximum value of the progress. Must be greater than `value`. The default is 100. */
+  max?: number;
+  /* Whether the progress should be split into individual steps. If so, `max` and value` are treated as the total and current steps respectively. */
+  showSteps?: boolean;
+  /* Whether the progress is inverted in color or not. */
+  inverted?: boolean;
+  /* [skip docs] */
+  'aria-label'?: string;
```
blurbyte, line 25: I think it's not really intuitive to show / hide label based on whether `aria-label` is provided or not. Unfortunately most devs don't really know what `aria-label` is and when to use it.
Not really sure what the best solution should be here. Looks like most design systems support 2 props to handle it:
- `hideLabel` - label is not visible, `aria-label` is applied
- `customLabel` (this one has different name everywhere) - well, your custom label, you can put there anything you want

That way we can automatically associate progress and label IDs, so user doesn't have to worry about this low level detail.

Now the question is how to handle scenario when both `customLabel` and `hideLabel` are provided. I think in this scenario we just grab text content of `customLabel` and put it as `aria-label`.

Also there is a question from where we take `value` for custom label. There a few possibilities / ideas:

- since it's controlled component anyway state is managed in a parent and passed to custom label directly
- could be via `const { value } = useProgress()` hook
- or via render prop `<Progress  customLabel={(value) => (<span>Label: {value}</span>) }/>`

src/progress/steps.tsx
```diff
@@ -0,0 +1,41 @@
...
+import Progress from './progress';
+
+type ProgressProps = {
+  id: string;
+  max: number;
+  size: NonNullable<React.ComponentProps<typeof Progress>['size']>;
+};
+
+function ProgressSteps({ id, max, size }: ProgressProps) {
+  function getStepClipRect(index: number) {
```
blurbyte, line 11: To follow convention maybe `renderStepRect`? Alternatively, could be separate `Step` component.

src/progress/progress.tsx
```diff
@@ -0,0 +1,63 @@
...
+  /* Optional identifier. Useful for when using your own `<label>`. */
+  id?: string;
+  /* The size of the progress. */
+  size?: 'small' | 'medium';
+  /* The current value of the progress. This is in relation to the `max` property. */
+  value?: number;
+  /* The maximum value of the progress. Must be greater than `value`. The default is 100. */
+  max?: number;
+  /* Whether the progress should be split into individual steps. If so, `max` and value` are treated as the total and current steps respectively. */
+  showSteps?: boolean;
```
blurbyte, line 21: The problem is that we have a lot of case like that:
Screenshot 2022-08-10 at 14 12 19

We have to handle a few scenarios:
1. regular progress bar without steps - value driven; default label is `x%`
2. progress bar with complete / full steps only - steps driven (or maybe value, but rounded one so for example if we have 3 steps and provided value 50 step 1 & 2 will be filled); default label is `x / y`
3. progress bar with step only partially filled (where it's only visual overlay but real value is displayed) - value driven; default label is `x%`

Not really sure what would be best API here. Probably we will want `steps` prop when we can specify how many we want to show. And after that most likely additional prop to control how they are handled: `stepsMode: 'showValue' | 'wholeSteps'` (no idea yet how to call it). We do something similar for `FormField` and `required`.

src/progress/styles.ts
```diff
@@ -0,0 +1,127 @@
...
+  `;
+}
+
+export function stepClipWrapperStyle() {
+  return css`
+    position: absolute;
+    display: flex;
+    flex-wrap: nowrap;
+    width: 100%;
+    height: 1px;
```
blurbyte, line 44: Any reason why it's 1px? Should't it be 4 / 8 px depending on size?

blurbyte (changes requested):
Amazing job, glad it's now responsive 🎉 . There are a few (major) improvements we can make to handle some common scenarios. We can sync tomorrow if needed.

src/progress/styles.ts
ixTec: So the svg itself doesn't need a height because it doesn't render anything directly. The clip-path has its own determined height. Unfortunately, Safari _doesn't_ do this, and requires the svg to have a minimum height/width (hence the 1px) 

I'll add a comment to this line to explain the above too

src/progress/progress.tsx
ixTec: I like the idea of a hook for getting the value for custom labels (passing the responsibility off to the consumer to add it if they want it in the label", as well as automatically linking the label to the progress via `aria-*`. Will implement this all 👍

src/progress/progress.tsx
blurbyte: This makes sense. The only thing which might be not intuitive at the moment is the behavior of `value` and `max` (but it's well documented). Still probably it's better that way rather than providing yet another prop (`steps`). We should probably mention in docs it's possible to provide decimal fractions to have a step partially filled.

As for label we should deffo round it. Not entirely sure of the behavior here. For values like 1.3/3 rounding it up to 2 makes sense, but for 2.6/3 rounding it up to 3/3 could be kind of weird (?), as it implies progress is complete:
Screenshot 2022-08-11 at 09 28 25

I'm not entirely sold if `showSteps` flag. Maybe it would be better to make it sth like `mode: 'steps' | 'continuous'`.

src/progress/progress.tsx
blurbyte: Checked briefly `aria-label` best practices for progress (couldn't find anything definitive), but looks like it doesn't have to announce progress (there are better aria attrs to do so). So it should simplify that part.

ixTec: Are we still wanting the hook for the value (label value)? If there's no requirement to provide the value via the aria label or actual label (and instead using `aria-value* attributes`, we may not need it?

src/progress/progress.tsx
ixTec: I think we probably want to stick to rounding down, given that for steps especially as example 1.3/3 is not 2 steps completed, it's one step complete and the second still in progress.

For `showSteps`, I'm unsure on what else we could refer to it as besides a `mode` prop with values of `"steps" | "single"` perhaps?

Agree with adding the `aria-value*` attributes, given how unreliable screen readers can be.

src/progress/progress.tsx
blurbyte: I think we can assume, since it's controlled component, value will come somewhere from parent. Most likely we don't have to do anything.

src/progress/progress.tsx
blurbyte: Yes, I think `mode` is better than flag. Just thinking ahead, in future probably we will have `variant: 'linear' | 'circular'` so `mode: 'steps' | 'single'` makes sense. Just maybe `continuous` or `uniform` would be slightly better than `single`.

ixTec: `mode` with `steps` and `continuous` is good with me

src/progress/progress.tsx
```diff
@@ -0,0 +1,79 @@
...
+          max={max}
+          value={value}
+          aria-label={ariaLabel}
+          aria-valuemin={0}
+          aria-valuenow={value}
+          aria-valuemax={max}
+          {...restProps}
+        />
+      </div>
+      {customLabel &&
```
blurbyte, line 68: Can't we just provide `label` element on our own? Than user won't have to worry about proper semantics (and we won't have to mention it in docs):
```
customLabel && <label htmlFor={progressId}>{customLabel}</label>
```

src/progress/progress.tsx
```diff
@@ -0,0 +1,79 @@
...
+type ProgressProps = {
+  /* Optional identifier. Useful for when using your own `<label>`. */
+  id?: string;
+  /* The size of the progress. */
+  size?: 'small' | 'medium';
+  /* The current value of the progress, which can also be a decimal. This is in relation to the `max` property. */
+  value?: number;
+  /* The maximum value of the progress. Must be greater than `value`. The default is 100. */
+  max?: number;
+  /* The mode in which the progress will display - `continuous` is the default behaviour, as one single bar and `steps` renders as multiple segments. For `steps`, `max` and `value` are treated as the total and current steps respectively. */
```
blurbyte, line 18: I would structure it a bit. We can use <br /> to make it more readable, something along the lines:
```
The mode in which the progress will display. <br/>
In default `continuous ` mode one single bar will be shown. <br/>
In `steps` mode multiple segments will be rendered. In this mode `max` and `value` are treated as the total and current steps respectively. Decimal value might be used as well to indicate step partially filled, e.g. `2.5` will show 2 and half bars filled. 
```

src/progress/progress.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { progressStyle, wrapperStyle, progressWrapperStyle } from './styles';
+import Steps from './steps';
+import Label from './label';
+
+type ProgressProps = {
+  /* Optional identifier. Useful for when using your own `<label>`. */
+  id?: string;
+  /* The size of the progress. */
+  size?: 'small' | 'medium';
+  /* The current value of the progress, which can also be a decimal. This is in relation to the `max` property. */
```
blurbyte, line 14: I would rather mention this decimals in `mode` prop docs, since it's most relevant for steps.

src/progress/progress.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <div data-testid="progress-wrapper" css={wrapperStyle({ size, inverted })}>
+      <div css={progressWrapperStyle()}>
+        {isStepsMode && <Steps id={progressId} max={max} size={size} />}
+        <progress
+          id={progressId}
+          css={progressStyle({ size, inverted, clipId: `${progressId}-clip` })}
+          max={max}
+          value={value}
+          aria-label={ariaLabel}
```
blurbyte, line 61: I think for screen reader user perspective it's better to announce steps before percentage value. We can enhance aria-label in similar way to what we did in Button with loader: so for steps it could be `ariaLabel.concat(' ', '${roundedVal} out of ${max} steps')`.

src/progress/label.tsx
```diff
@@ -0,0 +1,29 @@
...
+  max: NonNullable<React.ComponentProps<typeof Progress>['max']>;
+  value: NonNullable<React.ComponentProps<typeof Progress>['value']>;
+};
+
+function Label({ id, isStepsMode, max, value }: ProgressProps) {
+  const roundedVal = Math.floor(value);
+
+  return (
+    <label htmlFor={id}>
+      {isStepsMode ? (
```
blurbyte, line 17: Whether or not it's in steps mode we can easily hide in with `aria-hidden` (after we enhance `aria-label` with steps info).

src/progress/progress.tsx
```diff
@@ -0,0 +1,79 @@
...
+  id?: string;
+  /* The size of the progress. */
+  size?: 'small' | 'medium';
+  /* The current value of the progress, which can also be a decimal. This is in relation to the `max` property. */
+  value?: number;
+  /* The maximum value of the progress. Must be greater than `value`. The default is 100. */
+  max?: number;
+  /* The mode in which the progress will display - `continuous` is the default behaviour, as one single bar and `steps` renders as multiple segments. For `steps`, `max` and `value` are treated as the total and current steps respectively. */
+  mode?: 'continuous' | 'steps';
+  /* Custom label to be shown instead of the original. Use `<label>` element as the parent. */
```
blurbyte, line 20: I would suggest wrapping it in some label with basic styling so users don't have to worry about it.

blurbyte (changes requested):
We are getting really close to shipping it. Just a few a11y improvements.
As for `aria-label`:
- should be required by default
- should be optional when user provides their custom label, hard to predict what ppl will be putting there
- we don't really care whether label is hidden or not

src/progress/label.tsx
ixTec: Not sure I understand this comment, can you expand on it please?

src/progress/progress.tsx
ixTec: I think it's better to keep it on the `value` prop docs, since that's the actual prop it impacts. It also keeps the `mode` prop description a bit shorter too

src/progress/progress.tsx
ixTec: Edited the description here, but kept out the decimal bit as per https://github.com/datacamp/waffles/pull/77#discussion_r944342157

src/progress/label.tsx
blurbyte: Sure:

```
return (
    <label htmlFor={id}>
      <span aria-hidden="true">{isStepsMode ? `${roundedVal} out of ${max} steps` : `${roundedVal}/${max}`}</span>
    </label>
  );
  ```
  
But yeah, in this case probably it doesn't even have to be `label` with `for` attr (still ` aria-hidden` is required, even for div / span).

File: src/progress/progress.tsx (hunk 7 above)
Comment (blurbyte) [reply src/progress/progress.tsx:14]: Just trying to put myself in the user's shoes. It's fine to mention it here. But personally even after reading this I wouldn't make immediate connection I can use decimal values to make steps partially filled.

File: src/progress/label.tsx (hunk 9 above)
Comment (ixTec) [reply src/progress/label.tsx:17]: Ah gotcha, cheers

File: src/progress/progress.tsx [hunk 11]
```diff
@@ -0,0 +1,75 @@
...
+  /* The size of the progress. */
+  size?: 'small' | 'medium';
+  /* The current value of the progress, which can also be a decimal. This is in relation to the `max` property. */
+  value?: number;
+  /* The maximum value of the progress. Must be greater than `value`. The default is 100. */
+  max?: number;
+  /* The mode in which the progress will display. In default `continuous` mode one single bar will be shown. In `steps` mode multiple segments will be rendered and `max` and `value` are treated as the total and current steps respectively. */
+  mode?: 'continuous' | 'steps';
+  /* Custom label content to be shown instead of the original. Use `<label>` element as the parent. */
+  customLabel?: JSX.Element;
```
blurbyte, line 17: I think it might be good idea to allow string here, so `React.ReactNode | string` (since we're providing wrapper JSX.Element is not really that relevant here). Just provide basic styling for text (I guess we are doing it already).

blurbyte (changes requested):

src/progress/label.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const roundedVal = Math.floor(value);
+
+  return (
+    <label htmlFor={id}>
+      {customLabel ? (
+        customLabel
+      ) : (
+        <span aria-hidden="true">
+          {isStepsMode
+            ? `${roundedVal}/${max}`
```
blurbyte, line 23: Got a little feedback from designers. They suggested adding spaces around `/`, so for example: `3 / 5`.

blurbyte (changes requested):

src/progress/progress.tsx
```diff
@@ -0,0 +1,82 @@
...
+  customLabel?: React.ReactNode | string;
+  /* Whether the progress is inverted in color or not. */
+  inverted?: boolean;
+  /* [skip docs] */
+  hideLabel?: boolean;
+  /* Should always be provided with context of the progress. E.g. "Course Progress". */
+  'aria-label': string;
+} & React.HTMLAttributes<HTMLProgressElement>;
+
+type ProgressNoLabelProps = {
```
blurbyte, line 26: Imho not allowing to hide custom label is unnecessary limitation, I would drop it. If someone wants to do it for whatever reason, not sure why we shouldn't allow it?
So at the end ideally following cases should be handled via typings:

- with just default label `aria-label` is required
- if `customLabel` is provided `aria-label` is optional
- when both `customLabel` and `hideLabel` are provided, `aria-label` is required

To make it a bit nicer we can call current `ProgressProps` `ProgressBaseProps`, and the 'final' union `ProgressProps`.

blurbyte (changes requested):
Just a few final suggestions (mainly about tests) before merging.

src/progress/__tests__/progress.e2e.ts
```diff
@@ -0,0 +1,21 @@
+describe('Progress', () => {
```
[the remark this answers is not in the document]
ixTec: I can see the same for the steps test (but again, shows fine whilst the test is occurring. As the component appears to display correctly in the browser during normal use, I suspect this could just be a bug with clip paths in cypress perhaps? Can probably leave it for now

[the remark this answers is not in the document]
blurbyte: Let's leave it for now. Soon we will bump Cypress to v10+ and after that if problem persists we might reinvestigate it.

src/progress/progress.tsx
```diff
@@ -45,7 +44,7 @@ function Progress({
   hideLabel,
   'aria-label': ariaLabel,
   ...restProps
-}: ProgressProps | ProgressCustomLabelProps | ProgressNoLabelProps) {
+}: ProgressBaseProps | ProgressNoLabelProps | ProgressCustomLabelProps) {
```
blurbyte, line 47: To shorten it here we can add `type ProgressProps = ProgressBaseProps | ProgressNoLabelProps | ProgressCustomLabelProps` above component.

src/progress/progress.tsx
```diff
@@ -19,21 +19,20 @@ type ProgressProps = {
...
 type ProgressNoLabelProps = {
   /* Whether to hide the default label or not. Not necessary if `customLabel` is provided. */
   hideLabel: boolean;
-  customLabel: never;
   'aria-label': string;
-} & ProgressProps;
+} & ProgressBaseProps;
 
 type ProgressCustomLabelProps = {
   customLabel?: JSX.Element;
```
blurbyte, line 33: I think we want type to match base one here: `React.ReactNode | string`?

src/progress/progress.tsx
```diff
@@ -19,21 +19,20 @@ type ProgressProps = {
   inverted?: boolean;
   /* [skip docs] */
   hideLabel?: boolean;
-  /* Should always be provided with context of the progress. E.g. "Course Progress". */
+  /* Should always be provided with context of the progress, e.g. "Course Progress". */
   'aria-label': string;
 } & React.HTMLAttributes<HTMLProgressElement>;
 
 type ProgressNoLabelProps = {
   /* Whether to hide the default label or not. Not necessary if `customLabel` is provided. */
```
blurbyte, line 27: Probably 'Whether to hide the default label or not.' is enough here.

blurbyte (approved):
Component looks amazing 🎉 . Left a few minor comments. Be free to merge afterwards.

src/progress/progress.tsx
ixTec: Oh, good catch! 😅

PR #78, opened 2022-08-09
[WF-641] More Menu placement options

blurbyte:
# Added more Menu placement options

### 🚨 Breaking Changes

- Added more top and bottom `Menu` placements: bottomLeft, bottomRight, topLeft, topRight. Default is the same as before changes (was called `bottom` which now is `bottomRight`).
  -  if defaults are used (`placement` prop is omitted) - no changes must be made, behavior is the same as before
  -  if  bottom / top `placement` is explicitly configured it should be checked whether it's expected behavior or not

### ❓ Misc Changes

- Bumped versions of all core & a few dev dependencies (besides react / cypress / jest / next related)

## Screenshot(s):
Screenshot 2022-08-09 at 13 59 40


ixTec (approved):
Can't say I can see any issues with this, so approved :)

PR #79, opened 2022-08-15
[WF-636] Basic Card component

blurbyte:
# Basic Card component

### 🚀 Major Changes

- Created new `Card` component. Pretty much what Old Waffles offered but with simplified API (no one was using elevations).
  - Can act as any element, and in case interactive ones focus state is supported.
  - Decorators at the top of card can be added via `headstone` slot.
  - Mouse over indicator can be disabled with `disableHoverEffect`.

### 🔧 Minor Changes

- Added `Multidoc` icon
- Removed thin drop shadow in `Menu` and `Toast` in favor of border to match `Card` style

### ❓ Misc Changes

- Use `Card` in docs, and adjust borders of custom components to match its style

~**Question**: Is it better to select variant with hover disabled / enabled via `disableHover` flag or `variant = 'plain' | 'elevated'` (or some similar names)?~

## Screenshot(s):
Screenshot 2022-08-15 at 14 21 46
Screenshot 2022-08-17 at 12 28 51


doc-site/examples/card/with-headstone.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <>
+          <Avatar content={<PythonLogomark />} />
+          <Avatar content={<SqlLogomark />} />
+          <Avatar content={<RLogomark />} />
+        </>
+      }
+    >
+      <div
+        css={css`
+          padding: ${tokens.spacing.medium};
```
ixTec, line 26: Not sure we need the padding on a parent element instead of the Card itself?

src/card/__tests__/card.e2e.ts
```diff
@@ -0,0 +1,30 @@
+describe('Card', () => {
```
ixTec, line 1: Let's add a story for a card with the hover effect disabled

src/card/__tests__/card.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+    );
+
+    const card = container.querySelector('div');
+
+    expect(card).toBeInTheDocument();
+  });
+
+  it('sets the data attribute on the card', () => {
+    const { getByTestId } = render(
+      <Card data-testid="taylors-card">
```
ixTec, line 32: What's the usefulness of this test? Just seems to be testing adding an additional custom prop to the element? (testing `restProps`?)

src/card/__tests__/card.spec.tsx
```diff
@@ -0,0 +1,86 @@
+import { render, fireEvent } from '@testing-library/react';
+
+import { Card } from '../index';
+
+describe('Card', () => {
```
ixTec, line 5: Let's add a test for with the hover effect disabled

src/card/card-internal.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { cardStyle, headstoneStyle } from './styles';
+
+import type { PolymorphicRef, PolymorphicComponentProps } from '../helpers';
+
+type CardBaseProps = {
+  /* The content of the card. */
+  children: React.ReactNode;
+  /* Content to show sticking out the top of the card, e.g. can be used for technologies. Could be [Avatar](/components/avatar) from Waffles (use default `medium` size) or a custom component. */
+  headstone?: React.ReactNode;
+  /* Disables mouse over visual indicator. */
```
ixTec, line 14: Would change this to "Disables `onHover` shadow effect"

src/card/card-internal.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useFocusRing } from '@react-aria/focus';
+
+import { cardStyle, headstoneStyle } from './styles';
+
+import type { PolymorphicRef, PolymorphicComponentProps } from '../helpers';
+
+type CardBaseProps = {
+  /* The content of the card. */
+  children: React.ReactNode;
+  /* Content to show sticking out the top of the card, e.g. can be used for technologies. Could be [Avatar](/components/avatar) from Waffles (use default `medium` size) or a custom component. */
```
ixTec, line 12: Change to "Content positioned along the top of the card. Could be [Avatar](/components/avatar) from Waffles (use default `medium` size) or a custom component".

I think the documentation examples show the technology use case well enough

src/card/styles.ts
```diff
@@ -0,0 +1,62 @@
...
+  isFocusVisible,
+  hasHeadstone,
+  disableHover,
+}: CardStyleOptions) {
+  return css`
+    display: block;
+    position: relative;
+    ${hasHeadstone &&
+    css`
+      padding-top: 20px;
```
ixTec, line 22: Linking back to what I mentioned around default padding being built into the card, we will need to accommodate the header too

ixTec (changes requested):
Few comments and changes requested inline, along with the following:

- I think sticking to a boolean-based prop for the hover effect is best, although I think `disableHoverEffect` would be a clearer name 
- For all our examples, we include some padding. It would be worth adding some padding to the `Card` component by default, which can be overwritten by the consumer if needed. I can't think of many helpful cases where we wouldn't want at least _some_ padding originally.
- Should we update components such as `Notification` / `Toast` to use this new `Card` internally? Perhaps even `Dialog`, `Drawer` etc could benefit from a stylised `Card` too?
- Are we offering an `inverted` variant to begin with too? Or as a future improvement to the card

doc-site/examples/card/with-headstone.tsx
blurbyte: Yes, otherwise headstone will be misaligned. Alternatively it's possible to set 28px top padding on the card itself, but I think solution with wrapper is a bit cleaner.

src/card/__tests__/card.e2e.ts
blurbyte: Cypress doesn't support CSS hover. I tried it but there is no way to trigger it.

src/card/__tests__/card.spec.tsx
blurbyte: It's common across across DC products to add some kind of tracking IDs to everything so yes, it's basically checks if `data-*` is passed properly.

src/card/card-internal.tsx
blurbyte: I'm wondering what will be clearer for user 'on hover' or 'mouse over', considering nowadays many front-end devs are not THAT familiar with CSS.

src/card/styles.ts
blurbyte: Tbh at first I had default padding baked in, but turns out most real implementations in apps are adding some kind of horizontal separator or image or w/e, so they would have to remove padding anyway. Overall I don't like it as it is but it 1:1 to what `Card` from Old Waffles is, so at least migration will be easier.

blurbyte:
For now I just basically ported Old Waffles card without `elevation` API, so migration should be easier. Ideally in future we will have more advanced version of the component which will handle at least most common use cases. Some of my thoughts regarding feedback:

-`disableHover` should be fine, especially with an example
- in reality most of cards in apps have some custom content and paddings would have to be zeroed-out anyway
- in terms of updating other components I think we should wait for final version of the card, for now it's basically just 2 tokens
- `inverted` sounds like good idea for future update

ixTec (commented):
I still think `disableHoverEffect` is a clearer name for the prop as I could see confusion in thinking it disables the hover event / interaction for the component - perhaps we can ask some other developers on this one?

Regarding the padding, it's probably worth still adding some default padding (factoring in the top padding for the headstone too). Something like small or medium spacing would likely be enough. If someone wants something incredibly different they can wrap their content and pad the wrapper or overwrite the card component, but ultimately I think trying to provide a generic standard for card use would be useful.

src/card/card-internal.tsx
ixTec: I'd stick with `on hover` for this case imo

ixTec:
I'm noticing some odd flickering on the box shadow when it appears on hover (kinda hard to see on the recording). Is this the intentional look or?


https://user-images.githubusercontent.com/9938773/185109902-0f8b90c6-4a63-482e-885e-13bc1812f240.mov

ixTec:
Haven't been able to find some cubic values that help the above issue, so I'm just going to approve this. If anyone else raises an issue with it after release, we'll know it's not just my monitors / eyes 😄

ixTec (approved):

PR #80, opened 2022-08-15
[WF-640, WF-642, WF-645] Minor fixes and changes

ixTec:
# Minor fixes and changes

[WF-640](https://datacamp.atlassian.net/browse/WF-640)
[WF-642](https://datacamp.atlassian.net/browse/WF-642)
[WF-645](https://datacamp.atlassian.net/browse/WF-645)

### 🚨 Breaking Changes

- Default button `type` property has been changed from the native `submit` to `button`

### 🔧 Minor Changes

- Fixed shrinking issue with avatar inside of a flex wrapper with long content
- Icons within SideNav Items will now have their default size set based on the item
    - E.g. `small` items will have `xsmall` icons and `medium` items will have `medium` sized icons


src/menu/item.tsx
```diff
@@ -31,6 +31,8 @@ type ItemBaseProps = {
   showNotificationDot?: boolean;
   /* Indicates that the item is currently active. */
   isActive?: boolean;
+  /* [skip docs] */
+  type?: 'button' | 'reset' | 'submit';
```
blurbyte, line 35: Same as regular `Button`.

src/tabs/tab-internal.tsx
```diff
@@ -17,6 +17,8 @@ type TabBaseProps = {
   isActive?: boolean;
   /* [skip docs] */
   inverted?: boolean;
+  /* [skip docs] */
+  type?: 'button' | 'reset' | 'submit';
```
blurbyte, line 21: Same as regular `Button`.

blurbyte (changes requested):
Good job, Avatar fix and small icon for SideNav.Item already look good.
There is a little problem with Button, because `a` could also accept type prop (which is rather rare case but still), so it's getting a little problematic. Left possible solution in comments.

src/button/button-internal.tsx
```diff
@@ -80,6 +83,7 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
       {...(isLoading && {
         disabled: true,
       })}
+      {...(Element === 'button' && { type: type })}
```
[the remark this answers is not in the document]
ixTec: Had a few issues with the original typings for the `type` prop, but this has been resolved now so I can use your first suggestion :)

blurbyte (approved):
Looks, good 🚀.

PR #81, opened 2022-08-18
Button `isLoading` content fix

ixTec:
# Fix issue with content being hidden incorrectly when button `isLoading`

### 🔧 Minor Changes

- Fix css selector order by wrapping content with css instead


blurbyte (changes requested):
If it's CSS specifity problem we can fix it in a bit more elegant way in `loaderWrapperStyle` instead of adding another wrapper.
```
&& {
  opacity: 1;
  height: auto;
 }
 ``` 
 That way we won't need `wrapperStyle` too.
 What do you think?

ixTec:
This still has the same issue, as the `& > *` selector gets higher in comparison to the `opacity: 1` css. I also think the wrapper is actually a cleaner approach now compared to how it was before, short of using `!important` here.

blurbyte (approved):

PR #86, opened 2022-08-24
[WF-646] Update typings for notification and dialog internal buttons

ixTec:
# Update typings for notification and dialog internal buttons

## [WF-646](https://datacamp.atlassian.net/browse/WF-646)

These changes are in an effort to start moving towards a global-based approach to types, where component types are structured from shared types where possible. Future changes will likely involve shared types for things such as icons.

This should hopefully allow for a _one change updates everything_ dynamic when updating button types going forward.

### 🚀 Major Changes

- Typings for `Notification.ActionButton`, `Menu.Button` and `Dialog.Button` have been updated to be based on Waffles `Button` type.
    - Some props have been omitted to reflect requirements for these internal buttons, but the docs have been updated to show this.

### ❓ Misc Changes

- Updated the logic for extracting types for the docs, as `extract-react-types-loader` does not support `Omit<>` (among other things)
    - This does raise a bigger issue with this utility, and a stronger argument for upgrading to a different type-extraction / doc-generation in the future - something for future plans.

## Screenshot(s):
Screenshot 2022-08-24 at 09 20 36


src/dialog/button-internal.tsx
```diff
@@ -4,32 +4,15 @@ import { Button } from '../button';
...
-  fullWidth?: boolean;
+type DialogButtonProps = {
   /* Focus this particular button when dialog is opened. */
   /* @default false */
   autoFocus?: boolean;
-} & Omit<
-  React.ButtonHTMLAttributes<HTMLButtonElement>,
-  'autoFocus' | 'children'
->;
+} & Omit<React.ComponentProps<typeof Button>, 'autoFocus'>;
```
blurbyte, line 11: I think I did it that way at first but due to that polymorphic shenanigans it's 'loosing' these typings. You can check it in some of the docs or e2e tests examples that autocompletion is not working properly. So I'm afraid at least for now it has to be added manually.

blurbyte (changes requested):
I think there is a little problem here. With new updated typings autocomplete is not working - it's because these polymorphic types. TS can't figure out what to do with them and `React.ComponentProps` is not enough here. Probably its possible to write similar method to extract these polymorphic types but not sure if it's worth the effort.
I agree at some point we probably should put all these shared types somewhere.
Also agree we should replace this loader too. The only benefit I though it had was extracting default props, but since it's not working, so no reason to keep it imho.

ixTec:
Managed to fix the autocomplete issue. There's still an issue with the components being polymorphic, which seems to because by the `as` prop from `Button` having the wrong type when extended :/ - this is the only thing holding this back now. 

Screenshot 2022-08-24 at 16 58 22

I'll try and look into it tomorrow afternoon if I get time

ixTec:
Closing for now, in favour of https://github.com/datacamp/waffles/pull/88.

Typings refactor / overhaul to be handled at a later date

PR #87, opened 2022-08-24
[WF-637] Code & Link improvements

blurbyte:
# Make Link and Code more consistent

## [WF-637](https://datacamp.atlassian.net/browse/WF-637)

A little overhaul of `Code`, `CodeBlock`, and `Link` components so they behave in more consistent way (and stay in line with `Paragraph`, which is the most common use case). ~Some of these changes might be considered 'breaking' (in some rare scenarios) - have to think what exactly mark as breaking here.~

### 🚨 Breaking Changes

- `Code` by default inherits font size from a parent. _No need to set it manually in most cases._
- Icons within link have their `size` set automatically based on font size. _It's still possible to set custom size._
- Changed `Link` display to inline so it wraps nicer between the lines

### 🚀 Major Changes

- Updated `Code` component
  - Added `small` size (to match `Paragraph` API)
- Updated `CodeBlock` component
  - Added `small` size (to be consistent with `Code`)
- Updated `Link` component
  - By default it inherits font size (as it was)
  - Added `size` to match Paragraph / Code API
  - Animated underline

## Screenshot(s):
Screenshot 2022-08-24 at 15 57 26
Screenshot 2022-08-24 at 15 55 13
Screenshot 2022-08-24 at 15 55 22


doc-site/pages/components/link.mdx
```diff
@@ -10,15 +12,21 @@ import Imports from '../../components/imports';
...
 
 <Layout title="Link" description="Used for navigational actions that appear within or directly following a sentence.">
 
 ## Usage Guidance
 
 For standalone navigation actions use plain `Button` instead.
 
+<Notification
+  title="Default Size"
+  description="In general just let the link inherit font size from the surrounding text."
```
ixTec, line 27: I am not sure these kind of comments need to be within a `Notification` - probably best to just have it as a paragraph / bullet point under "Usage Guidelines"

src/code/__tests__/code-explicit-size.story.tsx
```diff
@@ -0,0 +1,39 @@
...
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+        tempor incididunt ut labore et dolore magna aliqua. Ut{' '}
+        <Code size="small">inline code fragment of the small size</Code> enim ad
+        minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip
+        ex ea commodo consequat.
+      </Paragraph>
+      <Paragraph>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+        tempor incididunt ut labore et dolore magna aliqua. Ut{' '}
+        <Code size="medium">inline code fragment of the small size</Code> enim
```
ixTec, line 24: "of the  _medium_ size"

src/code/__tests__/code-explicit-size.story.tsx
```diff
@@ -0,0 +1,39 @@
...
+        ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+        aliquip ex ea commodo consequat.
+      </Paragraph>
+      <Paragraph>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+        tempor incididunt ut labore et dolore magna aliqua. Ut{' '}
+        <Code size="large">inline code fragment of the small size</Code> enim ad
```
ixTec, line 31: "of the _large_ size"

src/code/styles.ts
```diff
@@ -6,18 +6,24 @@ import { hexToRgba } from '../helpers';
...
+  inherit: {
+    fontSize: '86%',
+  },
 };
 
 const baseCodeStyle = css`
   font-family: ${tokens.fontFamilies.mono};
   font-weight: ${tokens.fontWeights.regular};
-  line-height: ${tokens.lineHeights.relaxed};
+  line-height: inherit;
```
ixTec, line 26: We probably only want to inherit `line-height` when the `size` = `inherit` right? Just in case there's anything wild being set on the parent

doc-site/examples/code-block/sizes.tsx
```diff
@@ -21,6 +21,12 @@ function Example() {
 --flavor=woff
 --with-zopfli
 --layout-features="kern,liga,clig,ccmp"
+--unicodes="*"`}</CodeBlock>
+      <CodeBlock size="small">{`pyftsubset JetBrainsMono.ttf
```
ixTec, line 25: We probably need to add `flex-wrap` on a parent or something for the codeblock examples as they currently cause the container to be wider than it should (causing overflow on smaller screen widths)

doc-site/examples/code/basic-usage.tsx
```diff
@@ -0,0 +1,34 @@
...
+        You can install Waffles components library by running{' '}
+        <Code>npm install @datacamp/waffles</Code> with npm or{' '}
+        <Code>yarn add @datacamp/waffles</Code> with yarn.
+      </Paragraph>
+      <Paragraph>
+        You can install Waffles components library by running{' '}
+        <Code>npm install @datacamp/waffles</Code> with npm or{' '}
+        <Code>yarn add @datacamp/waffles</Code> with yarn.
+      </Paragraph>
+      <Paragraph size="small">
```
ixTec, line 25: I don't think we need `small` and `large` variants within our `basic-usage` example - just the default behaviour should be enough

ixTec (changes requested):
Looks great, love the new addition of the `small` variants for `Code` etc. Couple of comments/thoughts.

- We should update the badge/code for our default value in the props table to use this new variant (along with making badge small now). 
- `Link` prop docs for `iconLeft` and `iconRight` should be updated to remove the suggestion to use `small` icons with smaller text - this is now the default behaviour

In terms changes to mark as **🚨 Breaking Changes**, here are my thoughts:
- `Code` component now inherits  a `font-size: 86%` from parent by default (
    - Was `medium` `size` beforehand
- `Icon`s within `Link` components will now have their `size` set automatically based on `font-size`. This can be overwritten by manually setting a `size`.
- `Link` now has `display: inline` for it's default styling, so it wraps better between lines

doc-site/pages/components/link.mdx
blurbyte: Yeah, probably overkill here. Not even sure if it should be mentioned really :)

src/code/styles.ts
blurbyte: Makes sense.

doc-site/examples/code/basic-usage.tsx
blurbyte: Agree, especially now when they inherit font size doesn't make sense to show all sizes.

ixTec (changes requested):
very nearly there!

I would also add a new "Basic Usage" example for `Code Block` now too.

ixTec (approved):
LGTM :)

PR #88, opened 2022-08-25
[WF-646] Add support for `isLoading` on `Dialog` buttons

ixTec:
# Add support for `isLoading` on `Dialog` buttons

### 🔧 Minor Changes

- Add support for `isLoading` on `Dialog` internal buttons (and therefore also `AlertDialog` internal buttons too)

## Screenshot(s):
Screenshot 2022-08-25 at 15 23 31


blurbyte (approved):
Should do the trick. Would be nice to add a few _typo fixes_ here from previous PR.

PR #89, opened 2022-08-30
feat: public svgs

walze:
# Description

Added a symlink from the icons folder to the public folder, not sure if this is the best approach for this or if it would work but you guys can let me know how to do this better.

I tried importing the raw svg's from the `src` folder but i guess you don't ship them so they do not appear in `node_modules` folder.

# Explanation

Fetching the SVGs using requests, like so `http://localhost:3000/icons/regular/back.svg`, will allow us to lazy load them and involves less scripts being run which improves performance and SEO.

# Example
https://github.com/datacamp-engineering/next-marketing-app/blob/1af99d625ad80d536fe157bf52ad3aac04b73e52/components/SVG.tsx


walze:
> Hello, thanks for contribution! But, that's not really how you are suppose to use icons from Waffles. You don't want to import them from `src` but rather use them as React component, e.g. `<AddCircleIcon size="small" />`. More examples here: https://waffles-next.datacamp.com/components/icon/#examples. Or am I missing something?

Yes, we know that is how you're supposed to use them. I'm just arguing that, as an example, having them publicly served, will be better for performance/SEO/Web Vitals and for bundle sizes. Since Importing SVGs via JS comes with more scripts, which result in more load being transfered because of polyfills, libs, the raw code and the lack of lazy loading.

blurbyte:
Sorry but what you just written is not true at all, and it smells like premature optimization:

1. Waffles is 100% tree shakable so if you import let's say 1 icon (`import { AddCircle } from '@datacamp/waffles/icon'`) only this exact icon will be imported not whole icons bundle. 
2. In general 1 icon is only 150-300 bytes in size which it's negligible
4. If you really need to you can lazy load it with `React.lazy` since it's a component (not saying it's good idea)
3. >Since Importing SVGs via JS comes with more scripts, which result in more load being transfered because of polyfills, libs, the raw code
I assume you're using some kind of JS framework, so you have all of these, and icons are only very small part of it.

To me in this case readability / maintainability (since it's just a component) triumphs over debatable performance gains.

walze:
> 



> Sorry but what you just written is not true at all, and it smells like premature optimization:
> 
> 1. Waffles is 100% tree shakable so if you import let's say 1 icon (`import { AddCircle } from '@datacamp/waffles/icon'`) only this exact icon will be imported not whole icons bundle.
> 2. In general 1 icon is only 150-300 bytes in size which it's negligible
> 3. If you really need to you can lazy load it with `React.lazy` since it's a component (not saying it's good idea)
> 
> > Since Importing SVGs via JS comes with more scripts, which result in more load being transfered because of polyfills, libs, the raw code
> 
> I assume you're using some kind of JS framework, so you have all of these, and icons are only very small part of it.
> 
> To me in this case readability / maintainability (since it's just a component) triumphs over debatable performance gains.

We're using Next.js for some pages in Datacamp so `react.lazy` does not really work and not sure but `next/dynamic` might not do tree shaking. While I agree that the perf difference on an 1:1 basis is negligible, if you have 5 icons of around 1kb in the same page, it's 5 files that will need to be evaluated/executed/parsed through JS' slow engine it makes much more of a difference in Google's metrics. Having just 1 file that fetches and caches the svg would be much faster.

And honestly IMO doing `<LazyIcon name='clock' inverted size="small" />` looks much nicer than having an import.

PR #90, opened 2022-08-31
[WF-647] Add new `setTitleCase` helper

ixTec:
# Add new `setTitleCase` helper

### 🚨 Breaking Changes

- `Button` now has the `setTitleCase` helper applied to the child content by default, if it's of type string.
    - This can be disabled via the new `disableTitleCase` prop on `Button` (should be used sparingly).

### 🚀 Major Changes

- Added new `setTitleCase` helper for applying title casing to provided content
    - Used internally by `Button`

### 🔧 Minor Changes

- Fixed missing `hideLabel` entry in `Progress` component docs prop table
- Updated (hopefully) all headings across docs site, examples and tests to use newly aligned sentence casing

### ❓ Misc Changes

- Docs table of contents items utilise `disableTitleCase` prop, to solve issues with titles in camelCase
## Screenshot(s):
Screenshot 2022-08-31 at 16 29 07


ixTec:
@blurbyte 

Few comments on this initially for you - 

The hook does not work in the following cases:

- Words wrapped in brackets (e.g. "this is a (fun) button" becomes "This is a (fun) Button" (same applies for punctuation separating words (it works from spaces).
- Camel-cased words (obviously). I've added the `disableTitleCase` prop to button to get around this for things such as our table of contents items on docs.
- It will not maintain existing capitalisation. E.g. text with "click HERE for fun" will become "Click Here for Fun" etc.

We can solve these by adding further checks / configuration for the hook, but I wonder if it goes too far.

src/dialog/__tests__/__snapshots__/dialog.spec.tsx.snap
```diff
@@ -419,7 +419,7 @@ exports[`Dialog renders snapshot 1`] = `
       <span
         class="emotion-6"
       >
-        Dismiss
+         Dismiss
```
blurbyte, line 422: Looks like a space is added here which is kind of suspicious.

blurbyte (changes requested):
The more I think about it the more I am convinced it should be just helper function rather than hook.

Also I think we should change all our headings to _Sentence case_ since new guidelines were officially established on Slack :)

blurbyte (changes requested):
Looks great overall. Thanks for doing boring job and updating all headings.
I think only in global README it hasn't been updated yet.

doc-site/pages/components/helpers.mdx
```diff
@@ -62,6 +63,14 @@ Checks whether color is provided in **HEX** notation or not. Examples of **HEX**
 
 In a similar way to [React.Children.map](https://reactjs.org/docs/react-api.html#reactchildren), creates a new children array, but calls a provided function recursively on every child (and its children).
 
+## setTitleCase
+
+Converts the provided string to match the required title case. Waffles [Button](/components/button) uses this for provided string content by default.
```
blurbyte, line 68: Maybe `Converts provided text to _Title Case_. ...`

src/helpers/set-title-case.ts
```diff
@@ -0,0 +1,60 @@
...
+  'in',
+  'to',
+  'into',
+  'upon',
+  'near',
+  'with',
+  'of',
+];
+
+function setTitleCase(content: string) {
```
blurbyte, line 41: Missing some JSDocs here similar to other helpers.

blurbyte (changes requested):

src/button/button-internal.tsx
```diff
@@ -23,6 +25,9 @@ type ButtonBaseProps = {
   /* Sets the style of the button suitable for dark backgrounds. */
   /* @default false */
   inverted?: boolean;
+  /* Disables the default title casing for type `string` children. Use only in rare situations. */
```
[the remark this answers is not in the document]
ixTec: Good idea :)

blurbyte (approved):
Looks really nice, let's merge.

PR #92, opened 2022-09-02
[WF-566] Slider component

blurbyte:
# New Slider component

## Changes

Implemented new `Slider` component. Under the hood it uses `react-range` lib (only 5KB gzipped), some APIs and internal implementations are dictated by this lib.

Thoughts:

- Inputs handled in more SOLID than DRY way (I guess it's fine for readability)
- Might be better to have `value` to be `number | [number, number]` but it's kind of hard to sync it with lib, considering onChange callback, not really sure if's worth extra effort here
- For input width calculation some are hardcoded, and ideally probably should be taken from components itself, but not really sure if it's worth spending extra time here and complicating it even further

### 🚀 Major Changes

- Added `Slider` component:
  - it's controlled component so state is handled in the parent component
  - accepts array as a `value`: when one number is passed (e.g. `[10]`) single thumb / handle is shown, but for two numbers (e.g. `[10, 30]`) it shows range between two thumbs
  - min, max, and step (fraction could be set too) are configurable
  - all labels could be hidden with `hideLabels`
  - inputs to enter values directly could be displayed with `showInputs` (not affected by `hideLabels`), they are shown instead of current value(s)
  - error handling for inputs
  - inverted, disabled, focus styles are available
  - labels could be formatted with `formatLabel`, when e.g. currency must be selected
 
### ❓ Misc Changes

- Added `react-range` dependency to handle complex slider logic
- Bumped `floating-ui` version

## Screenshot(s):

When everything is fine:

slider
Some errors:

Screenshot 2022-09-20 at 11 40 14
Screenshot 2022-09-20 at 11 39 50


ixTec (changes requested):
Honestly looks great to the eyes, such a clean and easy user interaction!

Added a couple of comments / thoughts

- Inputs should likely just have a `min-width` instead of `width`, because there's some overflow issues with large number lengths 
    - Content should also probably be centre aligned?
    - Screenshot 2022-09-02 at 14 29 30
- I am _really_ not a fan of the term `Thumb`, and although I can see it's the name used by `react-range`, I would choose an alternative for our prop/naming for clarity, perhaps `Handle` or `Slider` itself, with the actual waffles component being renamed to `Interactive Range` or `Range Slider`. Something to discuss during work week perhaps

blurbyte:
Agree with `Thumb` - personally not a big fan of it too, but decided to it because `react-range` using it, and Material Design guidelines probably popularized this term. Will change it to `Handle`.
For inputs I think their widths will have to be calculated up front based on max / step values. It's super awkward if inputs change their size when typing (this would happen with `min-width`. Since values are numbers only it should be rather easy to implement (just digits count multiplied by some hardcoded value). I guess we shouldn't rally care about incorrect values here.

src/slider/inputs.tsx
```diff
@@ -0,0 +1,119 @@
...
+    ) {
+      onChange([value[0], parsedValue]);
+      setHasError([hasError[0], true]);
+    } else {
+      isWhitelisted(inputValue) && setHasError([hasError[0], true]);
+    }
+  }
+
+  function renderInput(
+    value: string | number,
```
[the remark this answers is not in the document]
blurbyte: Can't do that because, e.g. case when user want to type decimal `10.` it would have to be parsed to `10` immediately so `.` would be gone making whole experience very awkward.

src/slider/slider-context.tsx
```diff
@@ -0,0 +1,34 @@
...
+function SliderProvider({ children, ...value }: SliderProviderProps) {
+  return (
+    <SliderContext.Provider value={value}>{children}</SliderContext.Provider>
+  );
+}
+
+function useSlider() {
+  const context = useContext(SliderContext);
+
+  if (context === undefined) {
```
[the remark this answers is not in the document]
blurbyte: I guess it doesn't really hurt do have it there - could be easy to forget in future if we will decide to expose this hook.

src/slider/inputs.tsx
```diff
@@ -0,0 +1,119 @@
...
+  // Keep separate state for inputs, because some characters other than numeric must be covered
+  // Both input and error state is an array of one or two items, to mimic structure of Slider value
+  const [inputValue, setInputValue] = useState<Array<string | number>>(value);
+  const [hasError, setHasError] = useState(initalError(isSingleInput));
+
+  // Sync value from props and inputs
+  // Clear all inputs errors when Slider is dragged
+  useEffect(() => {
+    setInputValue(value);
+    setHasError(initalError(isSingleInput));
```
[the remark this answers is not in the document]
blurbyte: Time to invest in some spell checker plugin for VSCode :)

src/slider/limit-labels.tsx
```diff
@@ -0,0 +1,21 @@
+import { Text } from '../text';
+
+import { limitLabelsWrapperStyle, limitLabelStyle } from './styles';
+import { useSlider } from './slider-context';
+
+type LimitLabelsProps = {
+  formatLabel: (value: number) => number | string;
+};
+
+function LimitLabels({ formatLabel }: LimitLabelsProps) {
```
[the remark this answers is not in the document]
blurbyte: Was my initial thought for its name too, but was worried it would imply it has something to do with selecting custom range.

src/slider/slider.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <SliderProvider {...{ value, min, max, disabled, inverted }}>
+      <div>
+        {showInputs && (
+          <Inputs onChange={onChange} step={step} label={ariaLabel} />
+        )}
+        {!hideLabels && !showInputs && (
+          <ValueLabels formatLabel={formatLabel} />
+        )}
+        <Range
```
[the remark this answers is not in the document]
blurbyte: Good idea 👍

src/slider/styles.ts
```diff
@@ -0,0 +1,148 @@
...
+    height: ${tokens.sizing.small};
+    opacity: ${disabled ? tokens.opacity.high : 1};
+  `;
+}
+
+export function trackLineStyle() {
+  return css`
+    // Shift track line slightly so thumb aligns with labels nicer
+    width: calc(100% + ${THUMB_DOT_RADIUS}px);
+    margin-left: -${THUMB_DOT_RADIUS / 2}px;
```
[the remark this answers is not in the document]
blurbyte: I prefer to reserve `calc` for calculating dynamic values, instead of something I know during design.

src/slider/slider.tsx
```diff
@@ -0,0 +1,73 @@
...
+  max = 100,
+  step = 1,
+  disabled = false,
+  hideLabels = false,
+  showInputs = false,
+  inverted = false,
+  'aria-label': ariaLabel,
+  formatLabel = (value) => value,
+}: SliderProps) {
+  if (value.length > 2 || value.length === 0) {
```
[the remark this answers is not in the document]
blurbyte: Yeah, but it's kind of hard to make it compatible with `react-range` and `onChange` handler without some ugly TS casts. So not sure if it's worth the trouble.
Also not 100% convinced if it's good idea to offer single number or array depending on type of Slider you want to have. Especially for `onChange` you will have to remember to use the proper one again.

src/slider/inputs.tsx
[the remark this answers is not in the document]
ixTec: interesting issue - I hadn't considered this!

src/slider/slider.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { SliderProvider } from './slider-context';
+import RangeLabels from './range-labels';
+import Range from './range';
+import Inputs from './inputs';
+
+type SliderProps = {
+  /* Current slider value. Array of values. For single handle provide single value e.g. `[20]`, and for range provide two values, e.g. `[20, 50]`. */
+  value: number[];
+  /* Handler called when handle is moved. */
+  onChange: (value: number[]) => void;
```
ixTec, line 19: Really not sure we want onChange and onChangeEnd still - they're used for similar purposes internally, except within the react-range component. We likely just care to expose one of the two as a prop

ixTec (changes requested):
Looks like there's a handful of new issues and I've added some additional comments too:

- Input fields should probably show numbers to the the full decimal places, to avoid jumping between character lengths
- The error state still isn't very clear imo - perhaps we need design consultation because when an input is invalid, it gives no reason for why, including no indication for screen readers
- Story file names can probably be a bit more detailed e.g. `slider-with-inputs` instead of 'slider-input`

src/slider/slider.tsx
blurbyte: I think we need both:
- onChange is used to set a value (and is constantly updating when user moves a handle)
- onChangeEnd is used to retrieve the last value after handle stops moving
So the first one is not really feasible if you want for example send an XHR request with a slider value.

blurbyte: After discussion: we decided to remove this onChangeEnd callback.

src/slider/slider.tsx
```diff
@@ -0,0 +1,187 @@
...
+    ) {
+      onChange([parsedValue, value[1]]);
+      onChangeEnd?.([parsedValue, value[1]]);
+      setHasError([true, hasError[1]]);
+    } else {
+      isNotWhitelisted(inputValue) && setHasError([true, hasError[1]]);
+    }
+  }
+
+  function handleSecondInputChange(event: React.ChangeEvent<HTMLInputElement>) {
```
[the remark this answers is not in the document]
blurbyte: Moved them to `Input` component, should be easier to digest.

src/slider/slider.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const [hasError, setHasError] = useState(initialError(isSingleInput));
+
+  // Sync value from props and inputs
+  // Clear all inputs errors when Slider is dragged
+  useEffect(() => {
+    setInputValue(value);
+    setHasError(initialError(isSingleInput));
+  }, [value, isSingleInput]);
+
+  // Separate each input handler for readability
```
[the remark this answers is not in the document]
blurbyte: Added some additional comments.

src/slider/value-labels.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { valueLabelsWrapperStyle, valueLabelStyle } from './styles';
+import { useSlider } from './slider-context';
+
+type ValueLabelsProps = {
+  formatLabel: (value: number) => number | string;
+};
+
+function ValueLabels({ formatLabel }: ValueLabelsProps) {
+  const { value, inverted } = useSlider();
+  const isSingleValue = value.length === 1;
```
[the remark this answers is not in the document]
blurbyte: I would prefer to leave it at is is to be super explicit.

blurbyte:
A few improvements:
- added input error messages below Slider, they work nicely with screen reader
- moved each input handler from Slider to Inputs
- fixed an issue with an input when it was impossible to type value like `10.00001`

src/slider/inputs.tsx
```diff
@@ -0,0 +1,161 @@
...
+}: InputsProps) {
+  const { value, onChange, min, max, step, disabled, inverted } = useSlider();
+  const inputWidth = useMemo(
+    () => calculateInputWidth(min, max, step),
+    [min, max, step],
+  );
+
+  // Separate each input handler for readability
+
+  // Short overview of how each handler works:
```
ixTec, line 41: We could do with some more punctuation in these comments, just to help readability etc. See suggestion below:

``` 
  // 1. Only numeric values are allowed
  // 2. When user starts typing, any error for the Input is cleared
  // 3. For better UX, some additional checks are performed before setting final value on Slider. So sometimes Slider and Input values are out of sync, e.g. value ending with dot
  // 4. If Input value passes checks, parsed value is set on slider
  // 5. Otherwise an appropriate error is displayed
  ```

File: src/slider/inputs.tsx [hunk 13]
```diff
@@ -0,0 +1,161 @@
...
+
+    if (
+      isValidValue(targetValue, step) &&
+      parsedValue >= min &&
+      parsedValue <= value[1]
+    ) {
+      onChange([parsedValue, value[1]]);
+    } else if (isNotWhitelisted(targetValue)) {
+      if (parsedValue < min || parsedValue > value[1]) {
+        setError(['First input value is out of allowed range.', error[1]]);
```
ixTec, line 89: is _outside_ of

src/slider/inputs.tsx
```diff
@@ -0,0 +1,161 @@
...
+      } else {
+        setError(['First input value is not overlapping with step.', error[1]]);
```
ixTec, line 91: Maybe `First input value is a valid step increment` - if I'm understanding this error correctly

src/slider/inputs.tsx
```diff
@@ -0,0 +1,161 @@
...
+      parsedValue <= max
+    ) {
+      onChange([value[0], parsedValue]);
+    } else if (isNotWhitelisted(targetValue)) {
+      if (parsedValue < value[0] || parsedValue > max) {
+        setError([error[0], 'Second input value is out of allowed range.']);
+      } else {
+        setError([
+          error[0],
+          'Second input value is not overlapping with step.',
```
ixTec, line 117: Maybe `Second input value is a valid step increment` - if I'm understanding this error correctly

src/slider/inputs.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    if (
+      isValidValue(targetValue, step) &&
+      parsedValue >= value[0] &&
+      parsedValue <= max
+    ) {
+      onChange([value[0], parsedValue]);
+    } else if (isNotWhitelisted(targetValue)) {
+      if (parsedValue < value[0] || parsedValue > max) {
+        setError([error[0], 'Second input value is out of allowed range.']);
```
ixTec, line 113: is outside of

src/slider/slider.tsx
```diff
@@ -0,0 +1,117 @@
...
+import ValueLabels from './value-labels';
+import { initialError, hasError } from './utils';
+import { SliderProvider } from './slider-context';
+import RangeLabels from './range-labels';
+import Range from './range';
+import Inputs from './inputs';
+import Errors from './errors';
+
+type SliderProps = {
+  /* Current slider value. Array of values. For single handle provide single value e.g. `[20]`, and for range provide two values, e.g. `[20, 50]`. */
```
ixTec, line 14: Maybe reword as "Current slider value(s) as an array. For a single-value slider, provide a single value array e.g. `[20]`, and for a range slider provide two values, e.g. `[20, 50]`".

src/slider/slider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  value: number[];
+  /* Handler called when handle is moved. */
+  onChange: (value: number[]) => void;
+  /* The minimum allowed value of the slider. Can be decimal or negative. */
```
ixTec, line 18: "Can be _a_ decimal or..."

src/slider/slider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  /* @default 0 */
+  min?: number;
+  /* The maximum allowed value of the slider. Can be decimal or negative. */
```
ixTec, line 21: "Can be _a_ decimal or..."

src/slider/slider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  /* @default 100 */
+  max?: number;
+  /* The minimal distance between values. Can be decimal. */
```
ixTec, line 24: "Can be _a_ decimal."

src/slider/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    return message !== '';
+  });
+}
+
+// Determine allowed value for input
+export function isDecimal(value: string) {
+  return /^-?([0-9]{1,})?(\.)?([0-9]{1,})?$/.test(value);
+}
+
+// Check if value could be safely set on slider
```
ixTec, line 16: "..._can_ be..."

ixTec (changes requested):
I would say it's almost there!

Couple of minor comments in the code and the one below, before we're good to merge:

- Regarding the error messages for our inputs, we can likely set these strings to CONST's or util literals so we can just populate whether it's first/second input. this would remove the current duplication we have

We can definitely make some further improvements later on as a separate task :)

ixTec (approved):
Let's mergeeee 🥇

PR #97, opened 2022-09-23
Release Candidate - v1.15.0

ixTec:
👾


src/button/button-internal.tsx
```diff
@@ -73,6 +73,13 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
   }: ButtonProps<T>,
   ref?: PolymorphicRef<T>,
 ) {
+  // Handle cases when empty string is passed incorrectly
+  if (typeof children === 'string' && children.length === 0) {
+    throw new Error(
+      'Button cannot have an empty string for children. Use `<Button aria-label="button label" ... />` syntax instead.',
+    );
+  }
```
lensbart, line 81: This is very dangerous, as it can lead to runtime errors in production when upgrading (basically the same issue we faced, but with the error being thrown intentionally). Either there should be a check against `''` at the type level (not sure if easy to implement) so that all instances will be caught during development, or there should be an explicit check for `NODE_ENV` so that errors will only be thrown during development.

It’s less of a problem for new code, because it will probably be checked well enough. But I don’t think we can do this for code that will be upgraded.

```suggestion
  if (process.env.NODE_ENV === 'development' && typeof children === 'string' && children.length === 0) {
    throw new Error(
      'Button cannot have an empty string for children. Use `<Button aria-label="button label" ... />` syntax instead.',
    );
  }
```

lensbart: Even better:
```suggestion
  if (process.env.NODE_ENV === 'development' && typeof children === 'string' && children.length === 0) {
    console.warn(
      'Button cannot have an empty string for children. Use `<Button aria-label="button label" ... />` syntax instead.',
    );
  }
```

blurbyte: We will most likely rewrite it to use `console.error` so it won't be completely ignored (as for regular warning) and will be reported by Sentry. At the same time app won't crash.

blurbyte (approved):

PR #98, opened 2022-09-26
[WF-615] Update styling of documentation best practices cards

ixTec:
#  Update styling of documentation best practices cards

### ❓ Misc Changes

-  Updated styling of documentation best practices cards

## Screenshot(s):
Screenshot 2022-09-27 at 12 30 37


blurbyte (changes requested):
Looks really good overall. There is a small detail to update: borders are a bit too vivid right now, they should have some opacity (besides top one). I would suggest taking values from `Notification` (success and error variants). Because of it most likely some absolutely positioned 'decor' element will be needed (instead top border). It might justify styling custom card here instead of reusing our `Card` component.

blurbyte (approved):
Looks good, the transition between borders is slightly visible. If you are fancy be free to add this 'decor' element instead of border at the top. But can be done later - it's such a detail most ppl won't event notice notice.

ixTec:
I'm gonna say it's a minor thing and we can leave it for now 😄

PR #99, opened 2022-09-28
[WF-652] Improve icon-only Button and docs

ixTec:
# Improve icon-only Button and docs

### 🚨 Breaking Changes

- Throw error when incorrect syntax for `Button` is used with empty string children (`<Button ...>{''}</Button>` instead of `<Button ... />`
    - Technically a breaking change because this _will_ throw an error for all incorrect solutions

### 🔧 Minor Changes

- Updated prop documentation for `children` for `Button`

### ❓ Misc Changes

- Updated best practices for `Button`
- General grammar fixes for `Button` docs


src/button/button-internal.tsx
```diff
@@ -73,6 +73,13 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
   }: ButtonProps<T>,
   ref?: PolymorphicRef<T>,
 ) {
+  // Handle cases where incorrect syntax is used
```
blurbyte, line 76: I would make it a bit more precise 'When empty string is passed incorrectly'

src/button/button-internal.tsx
```diff
@@ -73,6 +73,13 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
+  if (typeof children === 'string' && children.length === 0) {
+    throw new Error(
+      'Button cannot have an empty string for children. Use `<Button ... />` syntax instead.',
```
blurbyte, line 79: Really like this error. Maybe would make sense to mention `aria-label` in `<Button ... />`?

doc-site/pages/components/button.mdx
```diff
@@ -17,36 +17,36 @@ import FullWidth from '../../examples/button/full-width';
 import AsALink from '../../examples/button/as-a-link';
 import Inverted from '../../examples/button/inverted';
 
-<Layout title="Button" description="A button triggers an event or action. It lets users know what will happen next. Use actionable copy, don’t be vague.">
+<Layout title="Button" description="A Button triggers an event or action. It lets users know what will happen next. Use actionable copy, don’t be vague.">
```
blurbyte, line 20: Not really a fan of capitalising Button on this page. Any particular reason for doing so?

blurbyte (changes requested):
A lot of nice updates. Any reason for capitalising 'button'?

doc-site/pages/components/button.mdx
ixTec: Since it's the name of our component, it should be capitalised. I've updated this throughout the Button docs where found

src/button/button-internal.tsx
ixTec: Good idea!

doc-site/pages/components/button.mdx
blurbyte: I still think it makes more sense to not capitalise it:
- we want our users to treat our components as a natural go-to elements, so whenever they want to place a button on the page they use `Button`, and I lower-case naming helps with assimilating this concept
- if we want to capitalise it, probably we should also render it as `Code`, which makes overall doc page too busy imho

ixTec: Hmm I think we have to be careful here when referring to one of _our_ components and a generic / native element. An example of this is icon, where we use it to primarily refer to our Icon component, not a custom SVG icon etc. Similar to our Input component, we don't want to confuse them with native HTML inputs. `Input` !== `input` etc.

I think we get away with not wrapping it as `Code`, and instead just linking off to the correct documentation page when not already on that page. E.g. references of `Icon` link off to Icon docs

blurbyte: For general guidelines I think speaking in more generic terms is better, so when we want the user to use that particular component (capitalised + link / code), like `Icon` or `Link` in this case, it will stand out more.

ixTec: Updated casing throughout button docs as discussed earlier today

doc-site/pages/components/button.mdx
```diff
@@ -23,77 +23,77 @@ import Inverted from '../../examples/button/inverted';
...
   </BestPractices.Recommended>
   <BestPractices.Discouraged>
 
-    - use a lot of buttons&mdash;too many calls to action cause user's confusion
-    - use multiple `primary` buttons when you have many actions at once
-    - use `primary` button to manipulate user
+    - use a lot of buttons&mdash;too many calls to action cause users confusion
+    - use multiple `primary` Buttons when you have many actions at once
+    - use `primary` Button to manipulate user
     - use `upgrade` button for actions other than upgrading to different plan
```
blurbyte, line 52: Should be capitalized here.

doc-site/pages/components/button.mdx
```diff
@@ -23,77 +23,77 @@ import Inverted from '../../examples/button/inverted';
...
   </BestPractices.Recommended>
   <BestPractices.Discouraged>
 
-    - overuse button with icon
+    - overuse Buttons with icon
     - provide both `leftIcon` and `rightIcon` at once&mdash;only allowed in rare circumstances
-    - provide both button label and `aria-label`&mdash;do it only in rare cases when additional information must be provided for users with disabilities
-    - for icon-only button specify `aria-label` for both button and icon components
+    - provide both Button label and `aria-label`&mdash;do it only in rare cases when additional information must be provided for accessibility
+    - for an icon-only button, provide `aria-label` for both `Button` and [Icon](/components/icon) components
```
blurbyte, line 96: Probably we can omit wrapping Button with code here (for BestPractices and PropsTable I think we can skip it).

src/button/button-internal.tsx
```diff
@@ -41,7 +41,7 @@ type ButtonIconOnlyProps = {
 
 type ButtonNoIconProps = {
   icon?: never;
-  /* The content inside the button. Most of the time should be a plain text. */
+  /* The content inside the button. Most of the time should be a plain text. Should never be an empty string. */
```
blurbyte, line 44: I think it should be capitalized here(?)

src/button/button-internal.tsx
```diff
@@ -10,28 +10,28 @@ import ButtonLoader from './loader';
...
   variant?: 'primary' | 'secondary' | 'plain' | 'destructive' | 'upgrade';
-  /* Defines the size of the button. In most cases, the default size should be used. */
+  /* Defines the size of the Button. In most cases, the default size should be used. */
   /* @default medium */
   size?: 'small' | 'medium' | 'large';
-  /* Whether the button is in it's loading state. The button will maintain the minimum width of it's child content. */
+  /* Whether the Button is in it's loading state. The Button will maintain the minimum width of it's child content. */
   /* @default false */
   isLoading?: boolean;
   /* Allows the button to grow to the width of its container. */
```
blurbyte, line 22: Should be capitalized here.

blurbyte (changes requested):
A few minor fixes, and we are good to go.

blurbyte (approved):
Looks nice 🚀

PR #101, opened 2022-09-29
[WF-656] Fix angle rendering of Loader

ixTec:
# Fix angle rendering of Loader

### 🔧 Minor Changes

- Refactored `Loader` structure and styling to more closely resemble original

**Comments:**

- I don't particularly like the naming for all one thousand wrapper classes in this component... but honestly I can't think of a convention more suitable so 🤷‍♂️


blurbyte (changes requested):
Looks nice, and it's working correctly in all browsers 🥇. For consistency sake let's use regular _functions_ in styles.

blurbyte (approved):
Let's merge 🚀!

PR #102, opened 2022-10-04
[WF-660] Fixed unique key prop warning

blurbyte:
# Fix unique key prop warning

## [WF-660](https://datacamp.atlassian.net/browse/WF-660)

Looks like `Each child in a list should have a unique "key" prop.` warning resurfaced again in MFE Header on _staging_ after bump to version of _floating-ui_ lib. It was popping for icon-only button.

### 🔧 Minor Changes
 
- Added key to each button icon, which seem to fix the warning.


ixTec (approved):
Looks good to merge🔑

PR #103, opened 2022-10-04
[WF-659] Fix Notification decor and icon z-index

blurbyte:
# Fix Notification decor and icon z-index

## Changes

`Notification` decorative element z-indexes were too high, and were causing problems with `Tooltip`.

### 🔧 Minor Changes

- Decrease z-index of `Notification` decor and icon

## Screenshot(s):

Before:
Screenshot 2022-10-04 at 12 00 26

After:
Screenshot 2022-10-04 at 12 08 38


hermansje:
Is a z-index needed at all?
I think the z-index can be removed from the icon and the icon background if the order of these divs is switched: https://github.com/datacamp/waffles/blob/c9d39d707a10bf6849964e0a596877e1310cef3b/src/notification-card/icon.tsx#L37-L38
If a z-index would be needed, it would only be needed on the first div and 1 would be enough.

blurbyte:
@hermansje Nice catch, you are totally right 👍.

src/notification-card/styles.ts
```diff
@@ -105,6 +105,7 @@ const invertedIconVariantMap = {
 export function iconWrapperStyle() {
   return css`
     position: relative;
+    z-index: ${tokens.zIndex.default};
     width: ${tokens.sizing.small};
```
ixTec, line 109: Do we even need to be setting a `z-index` for the icon wrapper here? If we're setting it to default, I think we're okay to just not have a z-index specified at all?

ixTec (changes requested):
Not sure we need to still have a z-index of the icon / icon wrapper

src/notification-card/styles.ts
blurbyte: I prefer to keep it just a tad higher than 0 (when it's unspecified), just to avoid some weird possible glitches in future.

ixTec (approved):

PR #104, opened 2022-10-04
[WF-657] Add `console.warn`s and `console.error`s to button

ixTec:
# Add `console.warn` and `console.error` support to Button

### 🚀 Major Changes

- `Button` now logs a `console.error` when used with incorrect empty-children icon-only syntax
    - This replaces the original `throw new Error` that was used
- `Button` now logs a `console.warn` when a redundant icon size is provided

### ❓ Misc Changes

- Added initial structure for debug messages from Waffles components
    - The aim is to use this going forward to provide warnings/errors to the browser for incorrect (but not breaking) implementations, or to recommend best practices going forward where typescript/eslint can't enforce them.

## Screenshot(s):
Screenshot 2022-10-04 at 16 29 07

### Notes:
Future work will include support for error-codes for referencing to the component docs for further information/explanation.


src/button/__tests__/button.spec.tsx
```diff
@@ -1,9 +1,18 @@
 import React, { useRef, useEffect } from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, waitFor } from '@testing-library/react';
 
 import { Button } from '../index';
 import { AddCircle, ChevronRight } from '../../icon';
 
+// These can probably be moved into a global location for all tests at a later date
```
blurbyte, line 7: Yes, I agree it would be very cumbersome to add it to each test manually.

src/helpers/log-waffles-debug.ts
```diff
@@ -0,0 +1,27 @@
...
+function wafflesDebugError(message: string) {
+  console.error(WAFFLES_ERROR_PREFIX.concat(message));
+}
+
+// Log a warning to the debug console
+function wafflesDebugWarning(message: string) {
+  console.warn(WAFFLES_WARN_PREFIX.concat(message));
+}
+
+// TODO: Add support for an error/warning code for association to docs - @ixTec
```
blurbyte, line 16: Really good idea.

src/helpers/log-waffles-debug.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Log a `console.warn` or `console.error` from within a Waffles component. Renders as a warning by default
+ *
+ * @param message The message to provide alongside the error/warning
+ * @param isError Specifies whether it should be logged as an error
+ */
+function logWafflesDebug(message: string, isError = false) {
```
blurbyte, line 23: Maybe it should be 2 separate functions: `logError` and `logWarning`?

src/button/debug-messages.ts
```diff
@@ -0,0 +1,7 @@
+// Constants to be used as content for the associated component's `console.error` or `console.warn`.
```
blurbyte, line 1: Don't think we have to put them in separate file yet. But in future with proper error codes we deffo will.

src/button/button-internal.tsx
```diff
@@ -73,17 +74,23 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
...
-      'Button cannot have an empty string for children. Use `<Button aria-label="button label" ... />` syntax instead.',
-    );
+    logWafflesDebug(MESSAGES.EMPTY_CHILDREN, true);
   }
 
   const Element = as || 'button';
   const { focusProps, isFocusVisible } = useFocusRing();
 
   function renderIcon(originalIcon: JSX.Element) {
+    // Log warning if Icon size is set unnecessarily to associated default
```
blurbyte, line 86: I like it as a proof of concept, but showing warning for such things is an overkill imho. I think things like that should be handled by linter or codemod (since they can be autofixed). Most devs ignore warnings anyway and using them for scenarios like that even deepens the problem (as in 'The Boy Who Cried Wolf' tale).
I would reserve warning in like 90% of cases to deprecation messages, which we will need in future.

src/button/__tests__/button.spec.tsx
```diff
@@ -346,3 +361,36 @@ describe('Button', () => {
     });
   });
 });
+
+describe('Button errors and warnings', () => {
```
blurbyte, line 365: Here 'errors and warnings' should be enough since test runner will append 'Button'.

blurbyte (changes requested):
Really good starting point for future development and a few nice ideas how to move forward.

src/helpers/log-waffles-debug.ts
ixTec: I had it like this originally, just for ease of usage where we might have multiple errors/warnings within one component but happy to change it back again too

src/button/debug-messages.ts
ixTec: I think it's much cleaner to keep the separate from the beginning, and also promotes good practice with them going forward!

src/button/button-internal.tsx
ixTec: I spoke briefly with @lensbart on the topic of linting for Waffles and it's certainly something I want to explore, but for the purpose of this work I think this is a good thing for us to show a warning for, due to how recent of a change it is (and therefore some people might still be setting the icon size unnecessarily). I think it's a good middle-ground until we provide some form linting or otherwise

src/button/__tests__/button.spec.tsx
ixTec: These are a separate block so have their own name associated to them

src/button/debug-messages.ts
blurbyte: If we want to go that path, shouldn't we put them in the some separate file in the root directory then? I imagine e.g. 'no empty children' error might be useful for other components too (especially keeping in mind we want error codes too in future).

ixTec: I think there are certainly some reusable content within messages, but since we just have them for button for now, we can expand upon this down the line

src/button/button-internal.tsx
blurbyte: I strongly disagree leaving this warning in this particular case. I think it's a bit of overreaction: going from 0 warning / errors to have them for every non important, non harmful thing, which on top of that could be auto fixed.
It's nice as a proof of concepts, but not really sure what's the problem / downside of leaving provided already default prop - it doesn't break anything / visually it's the same. If we will change the default size of an icon to `large` for example, then yes, it most likely should be communicated that way.

ixTec: Will remove this for now then and I can investigate linting etc as a future task

lensbart: I think it makes sense to only warn when the prop is being set as a simple string, e.g. `size="medium"`. I generally prefer `size={isAboveSmall ? 'medium' : 'large'}` to `size={isAboveSmall ? undefined : 'large'}`. So by necessity, it seems like this should be a linter rule (ideally, as Matt said, with autofix), because I don’t think we’ll be able to tell both cases apart in JavaScript.

src/button/debug-messages.ts
blurbyte: That's fine, not a big deal.

blurbyte (approved):
Looks good 🚀.

PR #106, opened 2022-10-10
[WF-664] Card focus within

blurbyte:
# Show hover effect when element inside is focused

### 🔧 Minor Changes

- Show drop shadow when element inside a `Card` is focused

## Screenshot(s):

https://user-images.githubusercontent.com/20565536/194825476-ed23e0d0-853d-4334-93cd-a1e54963bda6.mp4


src/card/styles.ts
```diff
@@ -42,9 +42,18 @@ export function cardStyle({
...
       &:hover {
         box-shadow: ${tokens.boxShadow.thick};
+        transform: translateY(-1px);
+
         ${isFocusVisible &&
         `box-shadow: 0 0 0 2px ${tokens.colors.blueDark}, ${tokens.boxShadow.thick};`}
-        transform: translateY(-1px);
+      }
+
+      &:focus-within {
```
ixTec, line 51: Looks like we've got some duplication here. I think we can simplify this whole block to be the following with the same effect
```
${!disableHoverEffect &&
  css`
    &:hover,
    &:focus-within {
      box-shadow: ${tokens.boxShadow.thick};
      transform: translateY(-1px);
  
      ${isFocusVisible &&
      `box-shadow: 0 0 0 2px ${tokens.colors.blueDark}, ${tokens.boxShadow.thick};`}
    }
`}
```

src/card/__tests__/card-basic.story.tsx
```diff
@@ -18,16 +19,13 @@ function Story() {
...
-            margin-bottom: 0;
-          `}
-        >
+        <Paragraph>
           Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
           eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad
           minim veniam, quis nostrud exercitation ullamco laboris nisi ut
           aliquip ex ea commodo consequat.
         </Paragraph>
+        <Button>Action</Button>
```
ixTec, line 28: I like the addition of a Button here. We should update our doc example with one to, so that the focus-within behaviour is shown

ixTec (changes requested):
functionally, works great! 😄 Just a couple of minor points to address to improve before merging

src/card/styles.ts
blurbyte: It works fine if card is `div` or `section`, but when it's a link it applies drop shadow to whole card when it's focused and we don't want that. So both selectors must be split because of it.

ixTec: Not sure I'm following with what you mean here

blurbyte: It's possible to set `<Card as="a" />` and thereafter if you focus a card, with combined selectors it applies hover drop shadow effect on top of blue border, where we want only border.

ixTec: ~Right, gotcha!

ixTec (approved):
Looks good!

PR #108, opened 2022-10-11
[WF-635] Add banner `mode` to Notification

ixTec:
# Add new banner `mode` to Notification

### 🚀 Major Changes

- Added `mode` prop to `Notification` component
    - defaults to `inline` which has no change in behaviour or appearance to existing Notification
    - `banner` value shows Notification as a top-banner style (see images below)
- Notification content now has max-width of 1200px

## Screenshot(s):
Screenshot 2022-10-11 at 13 36 13
Screenshot 2022-10-11 at 13 37 02


src/notification-card/notification-card-internal.tsx
```diff
@@ -23,16 +27,18 @@ function NotificationCardInternal(
   }: NotificationCardProps,
   ref?: React.Ref<HTMLDivElement>,
 ) {
+  const isIconCentered = closable && isContentCentered;
```
blurbyte, line 30: What exactly is the purpose of this var? Couldn't tell a difference on how notification rendered with the styles it was adjusting commented out.

src/notification/notification-internal.tsx
```diff
@@ -9,6 +9,9 @@ type NotificationProps = {
   title: string;
   /* Supportive content to display below the title. */
   description?: React.ReactNode;
+  /* Whether to display the notification as a banner or inline mode. */
```
blurbyte, line 12: Maybe 'Whether to display the notification as a banner or as an inline content.'?

blurbyte (changes requested):
Looks good, and seems to behave nicely on different resolutions.

Not really sure about `isIconCentered` var, doesn't seem to do anything.

I think we are missing a few tests here and there. Would be nice to have at least 1 jest snapshot of notification in banner mode.

Maybe it would be good idea to put some limit on how wide content could be (for `ContentContainer` it's 1200px for example). Here, albeit scenario is highly unlikely, it looks a bit weird:
Screenshot 2022-10-12 at 14 17 24

src/notification-card/notification-card-internal.tsx
ixTec: It's to add the required opposing margin when the notification is closable, which is now set as `margin-right: auto` so we need a `margin-left: auto` on the left most element as the Icon/heading/title aren't wrapped separate to the close button. See below screenshot for how it looks without:

Screenshot 2022-10-12 at 13 46 03

blurbyte: Ah, ok.

blurbyte (changes requested):
Nice improvements to tests!
We are close to merging it, but there are a few issues:

1. Tbh I would completely skip that width limit for inline mode (it's very unlikely usage scenario overall). Probably It looks better without it, and there is an issue with action button:

_Before_
Screenshot 2022-10-13 at 11 59 07

_Now_
Screenshot 2022-10-13 at 12 00 58

2. Width limitation in banner mode should align nicely with `ContentContainer`, which is not a case atm (would be ideal if left icons align nicely in scenario shown below):
Screenshot 2022-10-13 at 12 06 32

blurbyte (approved):
Looks good, let's merge 🚀.

PR #110, opened 2022-10-20
[WF-564] Table component

blurbyte:
# Table component, and indeterminate Checkbox

### 🚀 Major Changes

- Added new responsive `Table` component
  - Scrollable when content of the table doesn't fit in the viewport
  - Dark mode
  - Various utility components, which otherwise would be tricky to implement
    - `HeadCell` with sorting
    - `HeadCellCheckbox` and `CellCheckbox`
    - `CellMenu` (with `CellMenuCategory` and `CellMenuItem` re-exported from `Menu` for convenience)
- Added indeterminate `Checkbox` state

**TODO:**
- Use `Table` internally for props
- Add inline props and subcomponents to docs
- Investigate issue in docs when at certain breakpoint (around 850px) it doesn't want to shrink into container (might have something to do with grid layout)

## Screenshot(s):
Screenshot 2022-10-20 at 15 50 09
Screenshot 2022-10-20 at 15 50 20
t.com/20565536/196968238-db765006-a7b4-4326-9d96-b5dc5a3618f9.png">
Screenshot 2022-10-20 at 15 50 37


src/checkbox/__tests__/checkbox.e2e.ts
```diff
@@ -39,4 +48,9 @@ describe('Checkbox', () => {
     cy.get('main').findByText('Basic checkbox').click();
     cy.get('main').findByText('Basic checkbox').click();
   });
+
+  it("if indeterminate, when label is clicked doesn't change to checked", () => {
+    cy.loadStory('checkbox-indeterminate');
```
ixTec, line 53: Think this test is incomplete?  Just clicks it but doesn't check for anything after.

I also think we don't want this behaviour? A clicked indeterminate checkbox label should still toggle the checked state because an indeterminate can still have a functional checked/unchecked state (especially useful for forms). See https://baseweb.design/components/checkbox/ or https://ant.design/components/checkbox/ for good examples of this

src/checkbox/__tests__/checkbox.spec.tsx
```diff
@@ -92,6 +124,30 @@ describe('Checkbox', () => {
...
+        Taylor Swift
+      </Checkbox>,
+    );
+
+    const input = getByLabelText('Taylor Swift', { selector: 'input' });
+
+    expect(input).toHaveAttribute('aria-checked', 'mixed');
+  });
+
+  it("renders indeterminate checkbox, even when it's unchecked", () => {
```
ixTec, line 139: Same as previous comment. Perhaps simply renaming these tests to "renders indeterminate checkbox as unchecked/checked" for each of the two tests, although since they aren't rendered as snapshots I think it's bit  odd

src/checkbox/__tests__/checkbox.spec.tsx
```diff
@@ -100,73 +156,97 @@ describe('Checkbox', () => {
     expect(input).toHaveFocus();
   });
 
-  it('renders basic snapshot', () => {
-    const { container } = render(
-      <Checkbox onChange={jest.fn()}>Taylor Swift</Checkbox>,
-    );
+  describe('renders snapshot', () => {
+    it('with label and input', () => {
```
ixTec, line 160: We should have a snapshot of indeterminate checkbox too

src/checkbox/styles.ts
```diff
@@ -70,20 +79,30 @@ const checkmarkBaseStyle = css`
...
+  isIndeterminate: boolean,
+) {
   if (inverted) {
-    return checked ? tokens.colors.blue : tokens.colors.navyLight;
+    return checked || isIndeterminate
+      ? tokens.colors.blue
+      : tokens.colors.navyLight;
   }
-  return checked ? tokens.colors.blueDark : tokens.colors.white;
+  return checked || isIndeterminate
```
ixTec, line 97: empty line  after if statement before return

ixTec (changes requested):

src/checkbox/checkbox-internal.tsx
```diff
@@ -3,50 +3,80 @@ import { mergeProps } from '@react-aria/utils';
...
       />
-      <Checkmark {...{ inverted, checked, error, isFocusVisible }} />
-      <Text as="div" css={contentStyle({ inverted })}>
-        {children}
-      </Text>
-    </label>
+      {children && (
+        <Text
+          as="label"
+          htmlFor={checkboxId}
```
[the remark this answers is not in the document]
blurbyte: Should we do the same for switch / radio? Plus make it optional (with aria-label required).

src/checkbox/__tests__/checkbox.e2e.ts
blurbyte: I'm pretty sure indeterminate takes precedence over checked. You can check in advanced `Table` example how it works.

src/checkbox/__tests__/checkbox-indeterminate.story.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { Checkbox } from '../index';
+import { tokens } from '../../tokens';
+
+const wrapperStyle = css`
+  padding: ${tokens.spacing.medium};
+`;
+
+function Story() {
+  const [isChecked, setIsChecked] = useState(false);
```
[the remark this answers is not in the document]
blurbyte: It's used for checking / unchecking checkbox so it is required.

src/checkbox/__tests__/checkbox.spec.tsx
```diff
@@ -92,6 +124,30 @@ describe('Checkbox', () => {
     expect(input).toBeDisabled();
   });
 
+  it("renders indeterminate checkbox, even when it's checked", () => {
```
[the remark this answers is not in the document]
blurbyte: It tests whether indeterminate takes precedence over checked state (as it should).

src/checkbox/__tests__/checkbox.spec.tsx
blurbyte: just checking here whether checked / unchecked state affects indeterminate (and it shouldn't).

src/checkbox/__tests__/checkbox.spec.tsx
blurbyte: There is snapshot of it at the bottom.

src/checkbox/checkbox-internal.tsx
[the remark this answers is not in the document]
ixTec: Yep, sounds like a good idea to me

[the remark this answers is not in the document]
blurbyte: I will add a story to our board about it.

doc-site/components/markdown-elements.tsx
```diff
@@ -113,11 +113,26 @@ function CodeBlock({ children }: ContentProps) {
 }
 
 const tableStyle = css`
-  width: auto;
+  & th:last-of-type,
+  & td:last-of-type {
+    text-align: right;
```
ixTec, line 118: Hmm I don't know about setting this as the default table styling for markdown tables, since it's only useful for Roadmap. If we add any other tables in `.mdx` down the line, this will have to change. We should probably find a different approach for styling the roadmap table

ixTec (changes requested):
Haven't delved into the implementation of `Table` yet but I have a few comments around docs and examples etc:

- Docs props table doesn't seem to be properly responsive on smaller screen widths (<1200px). Page overflows instead of table scrolling, probably due to `min-width` applied for description
- Looks like best practices not recommended card now has a styling issue from these changes
Screenshot 2022-10-21 at 11 27 56
- Since this is a new component we plan to release, it's important we start releasing them alongside documentation. As a result, we should provide at the very least some initial usage guidelines and best practices for `Table` alongside this, instead of coming back and updating them later.

src/checkbox/__tests__/checkbox.e2e.ts
blurbyte: It is controlled component and doesn't contain any state / logic inside. Not really sure if there is any universal rule whether to show / hide this indeterminate checkmark. Would be great to see some possible implementation example, since I not really sure what are we trying to achieve by adding some extra logic / quasi state internally.

src/dialog/styles.ts
```diff
@@ -82,10 +79,12 @@ export function bodyStyle({
     flex-grow: 1;
     overflow-x: hidden;
     overflow-y: auto;
-    box-shadow: ${showShadowTop &&
-      'inset 0 12px 12px -12px rgba(5, 25, 45, 0.3)'}${showShadowTop &&
-      showShadowBottom &&
-      ','}${showShadowBottom && 'inset 0 -12px 12px -12px rgba(5, 25, 45, 0.3)'};
+    box-shadow: ${hasShadowTop &&
```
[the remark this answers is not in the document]
blurbyte: It's just renaming internal vars, not sure if it's worth documenting.

doc-site/examples/checkbox/states.tsx
```diff
@@ -12,6 +12,9 @@ function Example() {
         gap: ${tokens.spacing.large};
       `}
     >
+      <Checkbox isIndeterminate onChange={() => {}}>
+        Indeterminate checkbox
```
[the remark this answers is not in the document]
blurbyte: Good idea.

doc-site/components/best-practices-card.tsx
```diff
@@ -8,30 +8,49 @@ import {
...
-    border-top-color: ${color};
+    border-width: ${tokens.borderWidth.thin};
+    border-style: solid;
+    border-color: ${hexToRgba(
+      color,
+      isRecommended ? 0.8 : tokens.opacity.high,
+    )};
+    border-radius: ${tokens.borderRadius.medium};
+    padding: ${tokens.spacing.medium};
+    padding-top: 20px;
```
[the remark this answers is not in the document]
blurbyte: Got some style lint error so can't do it that way:
Screenshot 2022-10-21 at 13 56 13

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,63 @@
...
+import playgroundConfig from '../../examples/table/playground-config';
+import BasicUsage from '../../examples/table/basic-usage';
+import AdvancedUsage from '../../examples/table/advanced-usage';
+import Inverted from '../../examples/table/inverted';
+
+<Layout title="Table" description="🚧 This page is under construction">
+
+## Usage guidance
+
+<Notification
```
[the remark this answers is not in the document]
blurbyte: I'm not sure. It's probably most important thing ppl might want to know about this table.

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,63 @@
...
+  title="Customizing column width"
+  description={
+    <>
+      It is possible to fine tune column width by setting <Code>min-width</Code>{' '}
```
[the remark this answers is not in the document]
blurbyte: It's autoformatted that way with prettier.

doc-site/components/markdown-elements.tsx
blurbyte: As long it's the only MDX table not sure if it's worth spending time on it. If we will have more MDX tables happy to update it.

src/table/styles.ts
```diff
@@ -0,0 +1,305 @@
...
+    position: relative;
+    width: 100%;
+    transition: box-shadow 125ms ease-out;
+
+    ${isFocusVisible && `box-shadow: 0 0 0 2px ${tokens.colors.blueDark};`}
+
+    ${!hasShadowLeft &&
+    css`
+      border-top-left-radius: ${tokens.borderRadius.medium};
+      border-bottom-left-radius: ${tokens.borderRadius.medium};
```
ixTec, line 38: can be refactored to `border-radius: ${tokens.borderRadius.medium} 0 0 ${tokens.borderRadius.medium};`

src/table/styles.ts
```diff
@@ -0,0 +1,305 @@
...
+    `}
+
+    ${!hasShadowRight &&
+    css`
+      border-top-right-radius: ${tokens.borderRadius.medium};
+      border-bottom-right-radius: ${tokens.borderRadius.medium};
```
ixTec, line 44: can be refactored to `border-radius: 0 ${tokens.borderRadius.medium} ${tokens.borderRadius.medium} 0;`

src/table/__tests__/table.e2e.ts
```diff
@@ -0,0 +1,64 @@
+const screenSizes: Record<string, [number, number]> = {
+  small: [375, 800],
+  medium: [768, 1024],
+  large: [1920, 1080],
+};
+
+describe('Table', () => {
+  describe('when is longer than device width', () => {
```
ixTec, line 8: '...it's longer than...'
or
'...it is longer than...'

src/table/__tests__/table.e2e.ts
```diff
@@ -0,0 +1,64 @@
...
+      cy.get('main').findByLabelText('Regular table').focus();
+      cy.get('table').should('exist');
+    });
+
+    it("show shadow scroll hint to the right, when it hasn't been scrolled yet", () => {
+      cy.loadStory('table-regular');
+      cy.get('main').findByLabelText('Regular table').should('exist');
+    });
+
+    it('show shadow scroll hint on both sides, when it was scrolled to the middle', () => {
```
ixTec, line 24: we can scrap 'it was' from these titles of these tests

src/table/__tests__/table.e2e.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    it('show shadow scroll hint to the left, when it was scrolled to the end', () => {
+      cy.loadStory('table-regular');
+      cy.get('main').findByLabelText('Regular table').focus().scrollTo('right');
+      cy.get('table').should('exist');
+    });
+  });
+
+  // describe('render regular table with all cell types', () => {
```
ixTec, line 37: any reason these are commented out?

doc-site/examples/table/playground-config.ts
```diff
@@ -0,0 +1,47 @@
+import { Table } from '@datacamp/waffles/table';
+
+import type { PlaygroundConfig } from '../../types';
+
+const initialCode = `
+import { Table } from '@datacamp/waffles/table';
+
+function Playground() {
```
ixTec, line 8: I think playground should have at least one of the each cell type for ease of use when playing around component (e.g. a checkbox, menu etc)

src/checkbox/__tests__/checkbox.e2e.ts
ixTec: I'm definitely not trying to suggest we add some internal logic for updating the state for sure 😅 

It's a tricky situation because the indeterminate state of a checkbox should be largely visual-only. The underlying checkbox still has a checked value.

Right now if I have the following:
```
function Playground() {
  const [checked, setChecked] = useState(false);

  function toggle() {
    setChecked(!checked);
  }

  return (
    <Checkbox
      isIndeterminate
      checked={checked}
      onChange={toggle}
    >
      Checkbox description
    </Checkbox>
  );
}
```
And click the checkbox, the underlying checked value _will change_ to `checked=true`. However the `aria-checked` property will show `indeterminate`. 

There's inconsistency here, and from googling it looks like the checkbox should actually have the property of `indeterminate="true"` instead of the 'checked' property. See https://www.w3.org/TR/html-aria/#att-checked. This property can only be set via js, so will have to be set using a ref.

It also looks like the 'checked' property is calculated automatically from `value` and so we should be using `value` on our input instead of `checked`.

ixTec (changes requested):
Definitely getting closer to having this in top shape I think 💪 

- In hindsight, I think it would've been better to split the changes to Checkbox and Table into two separate PRs, to help with managing the scale of the PR/review.
- Given the complexity of this component, I think it would beneficial to split the directory up into some sub-folders, possibly a 'cells' folder with all cell and cell variant files, or a direct 'subcomponents' folder for all subcomponents instead. It's a minor project management thing but helps keep it a bit more readable when going through the file structure
- Scroll hooks can be refactored for a waffles hook using a similar approach to Tabs
- I agree with splitting up "Advanced Usage" example into individual sortable and checkbox examples

src/dialog/styles.ts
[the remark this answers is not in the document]
ixTec: I was initially confused as to what this change was, until I saw the terminology later on the the Table code

src/table/head-cell-checkbox.tsx
```diff
@@ -0,0 +1,23 @@
...
+  /* [skip-docs] */
+  isIndeterminate?: boolean;
+} & React.InputHTMLAttributes<HTMLInputElement>;
+
+function HeadCellCheckbox(props: HeadCellCheckboxProps) {
+  const { inverted } = useTable();
+
+  return (
+    <th css={headCellCheckboxStyle()}>
+      <Checkbox {...props} inverted={inverted} aria-label="Select all" />
```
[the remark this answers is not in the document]
blurbyte: Makes sense.

src/table/use-is-scrollable.ts
```diff
@@ -0,0 +1,34 @@
+import { useState, useEffect } from 'react';
+
+// Check whether wrapper is scrollable or not
+function useIsScrollable(
```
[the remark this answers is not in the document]
blurbyte: Yes, looks like it's pretty much exactly same functionality. I will move it to helpers, but not sure if it's worth documenting.

src/table/head-cell.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <button
+          {...focusProps}
+          onClick={onSort}
+          css={headCellSortButtonStyle({ isFocusVisible })}
+        >
+          {children}
+          {renderIcon()}
+        </button>
+      ) : (
+        children
```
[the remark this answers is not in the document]
blurbyte: It's extra HTML wrapper, as Chapeau is not polymorphic. As for icon we might actually need it, since saw something like this in our enterprise app:
Screenshot 2022-10-24 at 10 43 19

src/table/head-cell.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import { Sort, SortAscending, SortDescending } from '../icon';
+
+import { headCellStyle, headCellSortButtonStyle } from './styles';
+
+type HeadCellProps = {
+  /* Show appropriate sort icon, and sets `aria-sort` attribute. */
+  sort?: 'ascending' | 'descending' | 'indeterminate' | 'none';
+  /* Handler called when cell is clicked. */
+  onSort?: React.MouseEventHandler<HTMLButtonElement>;
```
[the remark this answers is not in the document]
blurbyte: Yes, probably regular `onClick` makes more sense here.

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+`Table.Body` accepts all props regular `HTMLTableSection` would, such as `data-` attributes.
+
+### Table.Row
+
+`Table.Row` accepts all props regular `HTMLTableRow` would, such as `data-` attributes.
+
+### Table.HeadCell
+
+In addition to all custom props listed below, `Table.HeadCell` accepts all props regular `HTMLTableCell` would, such as `data-` attributes. To customize column width it's possible to provide it via CSS `max-width`, `min-width`, or `width` (in this case percantages are recommended).
```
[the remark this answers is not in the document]
blurbyte: Wasn't sure if there will be any more in-depth docs in the first place 😄.

doc-site/examples/checkbox/indeterminate.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <div css={wrapperStyle}>
+      <Checkbox
+        isIndeterminate={isIndeterminate}
+        checked={allChecked}
+        onChange={(event) => {
+          setChecked(Array(2).fill(event.target.checked));
+        }}
+      >
+        Check all
```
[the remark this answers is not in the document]
blurbyte: Probably no need for need for `ScreenreaderOnly`. Label could be updated to be more explicit. Not sure if it needed here since element itself is announced as checkbox.

src/table/__tests__/table.e2e.ts
blurbyte: Forgot to uncomment these. Was playing with one of the tests above and these were pretty slow.

doc-site/examples/table/playground-config.ts
blurbyte: I think it is a bit too complex for playground, a lot of unusual handler, etc. which will make it less approachable.

blurbyte:
Since can't comment directly under the comment about indeterminate checkbox:

- Value is doing something different when type is set to checkbox as you can read [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#attr-value).
- All other design systems I checked are handling it in similar way as we do via `aria-checked` attribute
- What is most important, it's announced properly by screenreader

So I don't think we have to do any further refactors of internals.

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+### Table
+
+In addition to all custom props listed below, Table accepts all props regular `HTMLTable` would, such as `data-` attributes.
+
+<PropsTable
+  metadata={require('!!extract-react-types-loader!../../../src/table/table-internal')}
+/>
+
+### Table.Head
```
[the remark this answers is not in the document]
blurbyte: There is a lot going on, but I think we should _always_ keep such subcomponents as separate entries, even if they are identical. It is a matter of improving layout of this section visually. Maybe Maria will have some ideas how to do it. Although it's rather rare case when we have that many subcomponents.

doc-site/examples/table/playground-config.ts
ixTec: This is a fair point, we can keep it as is

src/table/use-is-scrollable.ts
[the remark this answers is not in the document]
ixTec: I think it could be helpful for other teams, doesn't hurt to offer alongside our other things

doc-site/pages/components/table.mdx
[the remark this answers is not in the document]
ixTec: I see that as being beneficial yeah, let's speak to Maria to discuss how best we can display this information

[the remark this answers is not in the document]
blurbyte: But don't think we have to do it right now, after release should be fine.

ixTec:
There's some really funky behaviour with different screenreaders and browsers it seems. w3 recommends not to set `aria-checked` when using `type=checkbox` and if it can be different to `checked`, but from testing, not all screenreaders recognise the `indeterminate` attribute for checkboxes... so I am unsure what is best here 😄 

I think we are okay with how you've implemented it, and can always adjust later on if needed

doc-site/components/best-practices-card.tsx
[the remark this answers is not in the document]
blurbyte: Ok, now I see why it wasn't working - I provided 4 values and when last one was moved to next line linter started to complain. Tbh not a big fan of _not_ specifying 2 / 4 in this shorthand notation, had to look up what is happening when you provide only 3.

blurbyte:
Pushed 1st batch of changes:

- allow setting custom `aria-label` on `HeadCellCheckbox` and `CellCheckbox`, tighten up typings too
- replaced `onSort` with regular `onClick`
- allow setting custom icon next to `HeadCell` label with `xsmall` size injected by default (as in `Tabs`)
- split advanced example into a few smaller ones
- various grammar / naming improvements

As for indeterminate checkbox I imagine a lot of screen readers are scanning only DOM and don't bother with JS-only attrs, so that's it might not work.

So tomorrow I am planning to:

- investigate buggy behavior of table + layout in docs
- refactor scrolling hooks
- add some guidelines

doc-site/components/best-practices-card.tsx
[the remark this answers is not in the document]
ixTec: can always specify 4 here, but I think it's good with the three values too

src/table/__tests__/table.e2e.ts
```diff
@@ -0,0 +1,64 @@
...
+      cy.viewport(screenSizes.medium[0], screenSizes.medium[1]);
+    });
+
+    it('is focusable', () => {
+      cy.loadStory('table-regular');
+      cy.get('main').findByLabelText('Regular table').focus();
+      cy.get('table').should('exist');
+    });
+
+    it('show shadow scroll hint to the right, when not scrolled yet', () => {
```
ixTec, line 19: We probably ought to test the mask/gradient in inverted mode too? Just to avoid any accessibility issues with future changes

src/table/__tests__/table.spec.tsx
```diff
@@ -0,0 +1,518 @@
...
+
+      const button = container.querySelector('button');
+      button && fireEvent.click(button);
+
+      expect(handleClick).toHaveBeenCalled();
+    });
+
+    it('when sort is not specified, and cell is clicked, appropriate handler is fired', () => {
+      const handleClick = jest.fn();
+      const { container } = render(
```
ixTec, line 203: I would have an empty line before our main render here

ixTec (changes requested):

blurbyte:
Pushed last batch of updates:
- added `useShowScrollHint` hook to handle shadows / masks for `Table`, `Dialog`, and `Tabs`; looks like 2nd ref wasn't even needed
- added basic docs to Table
- removed subcategories from Components in our side navigation, so it's just a flat list now
- some small examples tweaks

src/table/head-cell.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (sort === 'descending') {
+      return <SortDescending data-testid="sort-descending-icon" />;
+    }
+
+    if (sort === 'indeterminate') {
+      return <Sort data-testid="sort-icon" />;
+    }
+
+    if (icon) {
+      return icon.props.size ? icon : cloneElement(icon, { size: 'xsmall' });
```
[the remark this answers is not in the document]
blurbyte: Don't think we can do anything about it. It's more about how some icons look next to text so we would have to SVGs. Don't think there is any universal value we can put here which will fix it for all icons.

src/table/__tests__/table.e2e.ts
blurbyte: It is tested already. There are 2 describe blocks at the bottom just for that.

src/table/head-cell.tsx
```diff
@@ -0,0 +1,87 @@
...
+  sort = 'none',
+  icon,
+  onClick,
+  children,
+  ...restProps
+}: HeadCellProps) {
+  const { focusProps, isFocusVisible } = useFocusRing();
+  const isSortable = sort !== 'none';
+
+  function renderIcon() {
```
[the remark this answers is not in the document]
blurbyte: It was switch before. If we want to keep it default case will be rather complex. So pick your poison situation I guess.

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,154 @@
...
+  <BestPractices.Recommended>
+
+    - use to present tabular data, with a clear, repeating structure
+    - provide `aria-label` for Table
+
+  </BestPractices.Recommended>
+  <BestPractices.Discouraged>
+
+    - use for layout
+    - when there are fewer sections and each should be ephasized (use a [Card](/components/card))
```
ixTec, line 55: '... emphasized ...'

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+Keep in mind all components are _presentational_, and there is no built-in sorting.
+
+## Best practices
+
+<BestPractices>
+  <BestPractices.Recommended>
+
+    - use to present tabular data, with a clear, repeating structure
+    - provide `aria-label` for Table
```
ixTec, line 49: I would expand on this and say "provide `aria-label` for context of Table"

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,154 @@
...
+import WithMenu from '../../examples/table/with-menu';
+import Inverted from '../../examples/table/inverted';
+
+<Layout title="Table" description="Display and organize information from a data set.">
+
+## Usage guidance
+
+`Table` helps greatly with comparing, and analyzing data. It help users make informed decsisions. It should be used to present information with a clear, repeating structure.
+
+`Table` which is based on native HTML `table` and it's subcomponents provide a lot of structural information to screen reader users. People who relay on assistive technologies can easily navigate `Table` and identify relationships between data cells and columns. For that reason all of the `Table` components and subcomponents implement HTML `table` elements such as: `table`, `th` or `tr`.
```
ixTec, line 24: "... structural information **_for_** screen reader users."
"... people who _**rely**_ on assistive ..."

doc-site/pages/components/table.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+Using the `Table` in a project involves putting together following components and subcomponents:
+
+- `Table` is _responsive_ container, and `aria-label` must be provided to properly support scrolling the content with keybaord.
+- `Table.Head` and `Table.Body` are section elements which should contain only `Table.Row` subcomponents.
+- `Table.Row` is a single table row, and should wrap various cell subcomponents.
+- `Table.HeadCell` is used to label columns. Optionally provides sorting functionality. To customize column width it's possible to provide it via CSS `max-width`, `min-width`, or `width` (in this case percantages are recommended).
+- `Table.Cell` is regular data cell.
+
+There are also a few more advanced subcompnents which could be used in some niche situations:
```
ixTec, line 36: **_subcomponents_**

and we can omit "some" as "niche" already implies this

src/hooks/use-show-scroll-hint.tsx
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Occasionally for better user experience there is a need to hint that container is scrollable. Container should have `overflow: auto` or `overflow: scroll` specified via CSS.
+ *
+ * Depending on `orientation` and based on scroll position, show shadows or gradient masks at the appropriate positions.
+ *
+ * @param wrapperRef Ref of scrollable wrapper element
+ * @param orientation Whether to show hints to the left and right or top and bottom, either `horizontal` or `vertical`
+ * @returns Object of boolean values with `showStartHint`, and `showEndHint` entries
+ */
+function useShowScrollHint(
```
ixTec, line 25: We ought to have some tests for this new hook too

src/tabs/styles.ts
```diff
@@ -15,13 +15,11 @@ export function tabListStyle({
...
     margin-bottom: ${tokens.spacing.medium};
-
-    /* Make tabs scrollable, while hiding scroll, and showing appropriate gradient mask  */
-    mask-image: ${isRightGradientMaskVisible
-      ? isLeftGradientMaskVisible
-        ? `linear-gradient(90deg, rgba(0, 0, 0, 0), rgba(0, 0, 0, 1) ${tokens.spacing.xxlarge}, rgba(0, 0, 0, 1) calc(100% - ${tokens.spacing.xxlarge}), rgba(0, 0, 0, 0))`
-        : `linear-gradient(90deg, rgba(0, 0, 0, 1) calc(100% - ${tokens.spacing.xxlarge}), rgba(0, 0, 0, 0))`
-      : 'none'};
+    mask-image: ${isLeftGradientMaskVisible &&
+      `linear-gradient(90deg, rgba(0, 0, 0, 0), rgba(0, 0, 0, 1) ${tokens.spacing.xxlarge}, rgba(0, 0, 0, 1) calc(100% - ${tokens.spacing.xxlarge}), rgba(0, 0, 0, 0))`}${isLeftGradientMaskVisible &&
```
ixTec, line 19: Unsure if it's the styling changes here or the hook changes ,but looks like left gradient isn't being applied correctly for Tabs anymore

Screenshot 2022-10-26 at 09 43 17

doc-site/pages/components/hooks.mdx
```diff
@@ -50,6 +50,14 @@ Persists any function and prevents it from being invalidated between renders. Al
 
 Merge two React refs into single one in a stable way.
 
+## useShowScrollHint
+
+Occasionally for better user experience there is a need to hint that container is scrollable by either showing shadows or griadient masks. Depending on `orientation` (`horizontal` and `vertical` are available) and based on scroll position, show hints at the appropriate positions.
```
ixTec, line 55: _**gradient**_ masks

ixTec (changes requested):
I think we're very nearly there!

Some minor comments in the code again, and it looks like there are also some leftover comments that need resolving from previous review iterations!

src/hooks/use-show-scroll-hint.tsx
blurbyte: It's kind of tricky, since this hook is all about things you can't test in Jest. So it require spinning Cypress for it. Overall everything it provides is tested in Table and Dialog, so maybe it could be mentioned in comment?

src/tabs/styles.ts
blurbyte: Looks like it's styling issue.

src/hooks/use-show-scroll-hint.tsx
ixTec: I think this is a fair compromise yes

doc-site/examples/table/with-checkbox.tsx
```diff
@@ -0,0 +1,95 @@
...
+      return {
+        ...lang,
+        isChecked: event.target.checked,
+      };
+    });
+
+    setData(updatedData);
+  }
+
+  const areChecked = data.map((lang) => {
```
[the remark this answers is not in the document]
blurbyte: Resulting data structure is, e.g. [true, true, false] with each item representing checked state. Any idea how to better call it?

src/table/head-cell.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Sort, SortAscending, SortDescending } from '../icon';
+
+import {
+  headCellStyle,
+  headCellSortButtonStyle,
+  headCellWithIconStyle,
+} from './styles';
+
+type HeadCellProps = {
+  /* Show appropriate sort icon, and sets `aria-sort` attribute. Actual sorting action should be handled via regular `onClick`. */
```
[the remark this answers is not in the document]
blurbyte: I think 'icon' here is fine, since it's pretty generic UX pattern to show some kind of indicator column is sortable (be it some arrows, or something more fancy like here).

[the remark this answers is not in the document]
ixTec: Technically yeah, but given the icons we actually provide I think it's safe to use `Icon` here instead

src/table/head-cell.tsx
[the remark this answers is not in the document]
ixTec: I think we can clean up to have the `sort === *` comparisons to be a switch, and then keep our `if (icon)` as an if

doc-site/examples/table/with-checkbox.tsx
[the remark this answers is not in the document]
ixTec: I think `checkedMap` is good here, as it's a map of the checked state of the checkboxes

src/hooks/use-show-scroll-hint.tsx
blurbyte: I will add TODO comment here.

ixTec:
Looks like there's still a small issue with Tabs gradient on scroll, when there's no more overflow on the right:


https://user-images.githubusercontent.com/9938773/198037219-0fcc5b3f-358b-4d8a-b4b2-0a956b11ed6d.mov


We're _super_ close to being ready to merge otherwise. Just this one thing left

ixTec (approved):
Looks good to merge to me! Great job 🥇

PR #111, opened 2022-10-24
[WF-669] - Require `aria-label` property with Loader

ixTec:
# Require `aria-label` property with Loader

### 🚀 Major Changes

- Loader now requires `aria-label` prop

### 🔧 Minor Changes

- Fixed accessibility issue with Loader by adding a recommended `role=alert`
- ~~Due to weird behaviour with the polymorphic typings, I can't get `aria-label` to be required with `isLoading` for Buttons, so I've had to provide a `console.error` and a fallback `aria-label` for the Button's Loader to avoid this.~~
     - ~~Could probably be resolved if we move to a stricter typing style in the future (e.g. not polymorphic)~~


src/loader/loader.tsx
```diff
@@ -12,13 +12,24 @@ type LoaderProps = {
   /* Whether the Loader is inverted in color or not. */
   /* @default false */
   inverted?: boolean;
+  /* Should always be provided with context of the Loader. */
```
blurbyte, line 15: Maybe '...describing the context in which Loader is used?'

src/button/button-internal.tsx
```diff
@@ -100,7 +100,7 @@ function ButtonInternal<T extends React.ElementType = 'button'>(
         disabled: true,
       })}
       type={!type && Element === 'button' ? 'button' : type}
-      aria-label={isLoading ? ariaLabel?.concat(' ', 'Loading') : ariaLabel}
```
blurbyte, line 103: I think we most likely still need it because if it's icon-only button `aria-label` must be provided therefore it will 'overwrite' Loader aria-label.
That's another reason we don't want custom labels for loading. I think it's announced really nicely anyway (old label + loading).

doc-site/examples/button/states.tsx
```diff
@@ -13,7 +13,9 @@ function Example() {
     >
       <Button>No State</Button>
       <Button disabled>Disabled</Button>
-      <Button isLoading>Original Label</Button>
+      <Button isLoading aria-label="Loading Button">
```
blurbyte, line 16: We don't need `aria-label` here.

src/button/__tests__/button-variants-and-sizes.story.tsx
```diff
@@ -73,6 +73,7 @@ function Story() {
                 return (
                   <Button
                     isLoading
+                    aria-label={`Loading ${variant} ${size}`}
```
blurbyte, line 76: We don't need `aria-label` here.

src/button/__tests__/button-variants-and-sizes.story.tsx
```diff
@@ -150,6 +151,7 @@ function Story() {
                   <Button
                     isLoading
                     inverted
+                    aria-label={`Inverted loading ${variant} ${size}`}
```
blurbyte, line 154: We don't need `aria-label` here.

src/button/__tests__/button.spec.tsx
```diff
@@ -243,7 +243,9 @@ describe('Button', () => {
 
   it('renders loader with isLoading prop', () => {
     const { getByText, getByTestId } = render(
-      <Button isLoading>Add to Fanpage</Button>,
+      <Button isLoading aria-label="Adding to fanpage">
```
blurbyte, line 246: We don't need `aria-label` here.

src/button/__tests__/button.spec.tsx
```diff
@@ -334,7 +336,12 @@ describe('Button', () => {
       sizes.forEach((size) => {
         it(`variant ${variant} and size ${size}`, () => {
           const { container } = render(
-            <Button inverted variant={variant} size={size} isLoading>
+            <Button
+              variant={variant}
+              size={size}
+              isLoading
+              aria-label={`Loading ${variant} ${size}`}
```
blurbyte, line 343: We don't need `aria-label` here.

blurbyte (changes requested):
I think we still want to append 'Loading' to aria-label if it is provided, since screen reader doesn't announce it properly for icon-only button.
There are also some redundant `aria-labels` leftovers in tests and examples.

src/button/button-internal.tsx
ixTec: When button `isLoading` then it's disabled which means it's not focusable via keyboard navigation so technically a screen reader won't even know it's there as it ignores disabled buttons

blurbyte: Np!

blurbyte (approved):
Looks good 🚀

PR #113, opened 2022-10-28
[WF-675]  No label and required aria-label for Switch and Radio

ixTec:
# Add support for no label and required aria-label for Switch and Radio

## [WF-675](https://datacamp.atlassian.net/browse/WF-675)

**Comments:**

- Should we show a `console.error` on empty string as children, like do we do for Button?

### 🚀 Major Changes

-`children` is now optional for `Switch` and `Radio`, with `aria-label` required when it's not provided

### ❓ Misc Changes

- Removed obsolete `placeholder` prop on Checkbox example 
- Added no-label doc example for Checkbox

## Screenshot(s):
Screenshot 2022-10-28 at 14 15 56
Screenshot 2022-10-28 at 14 22 31
Screenshot 2022-10-28 at 14 22 47


doc-site/examples/checkbox/basic-usage.tsx
```diff
@@ -6,11 +6,7 @@ function Example() {
   const [checked, setChecked] = useState(false);
 
   return (
-    <Checkbox
-      checked={checked}
-      onChange={() => setChecked(!checked)}
-      placeholder="Welcome"
-    >
+    <Checkbox checked={checked} onChange={() => setChecked(!checked)}>
```
blurbyte, line 9: Nice catch.

doc-site/examples/checkbox/no-label.tsx
```diff
@@ -0,0 +1,16 @@
+import { useState } from 'react';
```
blurbyte, line 1: Thanks for adding tests and examples for `Checkbox`. Completely forgot about it when implementing intermediate flag.

doc-site/examples/switch/no-label.tsx
```diff
@@ -0,0 +1,23 @@
+import { useState } from 'react';
+import { css } from '@emotion/react';
+import { Switch } from '@datacamp/waffles/switch';
+
+function Example() {
+  const [checked, setChecked] = useState(false);
+
+  return (
+    <form
```
blurbyte, line 9: We don't need wrapper here, since there is no label.

blurbyte (changes requested):
Looks really good, and works like a charm.

Just a few minor things to polish a bit:
- for Switch and Radio I would add 1 more unit tests to check if `id` and `for` match (as in 'label and input are associated by the same ID' for Checkbox)
- for examples we don't need wrappers (could be unstyled `form` element too)
- I would move examples in docs to the bottom (just before inverted), since it's not a pattern we want to promote overall :)

blurbyte (approved):
Nice, let's merge 🚀.

PR #115, opened 2022-10-31
[WF-665] Allow single toast to be persisted 

ixTec:
# Add support for individual `disableAutoHide` prop on Toast

### 🚀 Major Changes

- Individual `Toast` can now have `disableAutoHide` provided, to override any value from `ToastProvider` parent


src/toast/toast-provider.tsx
```diff
@@ -14,32 +14,37 @@ type ToastsDataset = Record<
 >;
 
 type ToastProviderProps = {
-  /* Turns off toasts auto-hide functionality. When flag is passed, notifications persist and must be dismissed manually. */
+  /* Turns off Toast's auto-hide functionality. When the flag is passed, notifications persist and must be dismissed manually. */
```
blurbyte, line 17: I think it makes sense to keep it plural as it applies to all toasts.

src/toast/toast.tsx
```diff
@@ -7,45 +7,46 @@ import { animatedWrapperStyle } from './styles';
...
 type ToastProps = {
-  /* The main content of the notification. */
+  /* The main content of the Toast. */
   title: string;
   /* Supportive content to display below the title. */
   description?: React.ReactNode;
   /* Defines the type of notification. */
   /* @default default */
   variant?: 'default' | 'success' | 'warning' | 'error';
+  /* Overrides the `ToastProvider` value and turns off auto-hide functionality for the specific Toast. */
```
blurbyte, line 17: I would maybe rephrase it a bit to put more important bit first: 'Turns off auto-hide functionality for the specific Toast. Overrides the `ToastProvider` value'.

blurbyte (changes requested):
Well done 👍.

Would be nice to add 1 e2e story testing the behavior: 

- put there 2 buttons, one to trigger regular toasts, and 1 for the ones with auto hide disabled
- pop like 4 of them with different descriptions so it will be easy to identify in the next step
- after 6 seconds check if appropriate ones are still visible

blurbyte (approved):
Well done.

PR #117, opened 2022-11-03
[WF-693] Add `SkillTag` component

ixTec:
# Added new `SkillTag` component

### 🚀 Major Changes

- Added new `SkillTag` component

## Screenshot(s):
Screenshot 2022-11-03 at 15 01 49


doc-site/pages/components/skill-tag.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+import playgroundConfig from '../../examples/skill-tag/playground-config';
+import BasicUsage from '../../examples/skill-tag/basic-usage';
+import Variants from '../../examples/skill-tag/variants';
+import Levels from '../../examples/skill-tag/levels';
+import Inverted from '../../examples/skill-tag/inverted';
+
+<Layout title="Skill Tag" description="A small visual label used to show the skill level of a course or track.">
+
+## Usage guidance
```
blurbyte, line 18: I would make it a bit shorter, it's very simple and straightforward component at the end. Some rough edits:

1. A `SkillTag` can be used in course or track description cards and pages to indicate their skill level. It should never be used on its own, but rather to provide more visual information and catch user's attention.

2. Don't think we need 2nd paragraph at all. It's pretty much reiterating what we have in props.

3. Use 'variant=primary' when it's positioned near the top of the page or with an intent of catching the eye. In contrast, `variant='secondary'` should be used in cases where the `SkillTag` provides supportive information to the surrounding content.

src/skill-tag/__tests__/skill-tag-variants-and-levels.story.tsx
```diff
@@ -0,0 +1,65 @@
+import { css } from '@emotion/react';
```
blurbyte, line 1: I would split it into 2 e2e tests: `skill-tag-regular` and `skill-tag-inverted`.

src/skill-tag/level-indicators.tsx
```diff
@@ -0,0 +1,37 @@
+import { indicatorWrapper } from './styles';
+import SkillTag from './skill-tag';
+import Indicator from './indicator';
+
+const BEGINNER_COUNT = 1;
```
blurbyte, line 5: Probably it's cleaner to use map instead of consts + switch here:
```
const indicatorsCount = {
  beginner: 1,
  intermediate: 2,
  advanced: 3
}
```

and later to be consistent with what we have in `Progress` component:
```
Array.from({ length: indicatorsCount[variant] }, (_, index) => ...
```

src/skill-tag/skill-tag.tsx
```diff
@@ -0,0 +1,46 @@
...
+} & React.HTMLAttributes<HTMLDivElement>;
+
+function SkillTag({
+  variant = 'primary',
+  level = 'beginner',
+  inverted = false,
+  ...restProps
+}: SkillTagProps) {
+  // Obtain the associated label based on the skill level
+  function levelLabel() {
```
blurbyte, line 27: Maybe `renderLabel()`

blurbyte (changes requested):
Looks nice. 

I think _Usage guidance_ section could be a bit shorter considering how simple this component is. And currently it's mostly reiterating what we have in props table anyway.

Side navigation and component status entries are also missing.

As for `<ScreenreaderOnly>` it could be indeed good idea to add something like: `<Text>Beginner <ScreenReaderOnly>Skill Level</ScreenReaderOnly></Text>`.

src/skill-tag/skill-tag.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import { Text } from '../text';
+
+import { labelStyle, skillTagStyle } from './styles';
+import LevelIndicators from './level-indicators';
+
+type SkillTagProps = {
+  /* Determines the variant of the Skill Tag. */
```
blurbyte, line 9: Maybe _SkillTag_ without space?

doc-site/pages/components/skill-tag.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+When positioned near the top of a page, or used with the intent of catching the eye, then `variant='primary'` should be used. In contrast, `variant='secondary'` should be used in cases where the `SkillTag` is secondary to the surrounding content and not a focus.
+
+## Best practices
+
+<BestPractices>
+  <BestPractices.Recommended>
+
+    - use as alongside other content
+    - use the correct variant for the use case
```
blurbyte, line 32: I think it's pretty obvious and does't have to be stated here.

blurbyte (changes requested):

blurbyte (approved):
Nice new component for next release. Let's merge 🚀 !

PR #119, opened 2022-11-04
[WF-696] Add new ALPA icon assets

ixTec:
# Add new ALPA icon assets

## Comments

- Had to change the name for the alpa loop assets, in order to differentiate them properly from these new icons

### 🚨 Breaking Changes

- Renamed the original ALPA loop assets to be called `AlpaLoop` instead of `Alpa` (e.g.`ApplyColoredAlpa` => `ApplyColoredAlpaLoop`)

### 🚀 Major Changes

- Add new `Alpa` icon assets

## Screenshot(s):
Screenshot 2022-11-04 at 15 34 20


doc-site/pages/components/asset.mdx
```diff
@@ -57,9 +57,13 @@ An `Asset` should be used to provide visual context or expand upon a written poi
 
 <AssetGrid assetType="3d" />
 
+## ALPA Icon Assets
```
blurbyte, line 60: I think 'Icon' here is unnecessary here.

blurbyte (approved):
Should be good for now 👍 . Just not sure they should be called 'Icon', because they will be usually used as regular (pretty big) image.

PR #120, opened 2022-11-08
[WF-653] Design Principles docs page

blurbyte:
# Add Design Principles page to docs

### ❓ Misc Changes

- Added _Design Principles_ page to our docs

## Screenshot(s):

Screenshot 2022-11-08 at 13 41 05


ixTec (changes requested):
I think you're right in suggesting it should go in the foundation section instead.

I would also center-align the last card as well (when in a grid formation)

blurbyte:
I think it's fine to leave it left aligned. Not sure if's worth spending more time on polishng design of this page ;)

ixTec (approved):

PR #121, opened 2022-11-08
[WF-677] Add override for Notification close button

ixTec:
# Add override for Notification close button

### 🚀 Major Changes

- Exposed `Notification.CloseButton` to be used in the new `closeButtonOverride` prop in order to specify custom close button overrides


src/notification/notification-internal.tsx
```diff
@@ -36,6 +48,7 @@ function NotificationInternal({
   closable = false,
   onClose,
   action,
+  closeButton,
```
blurbyte, line 51: Maybe we should take it as an opportunity and introduce some basic `overrides` API? It would be basically a collection of predefined slots of internal components you would be able to replace with your custom ones. 

So in this case: `<Notification overrides={{ CloseButton: <CustomCloseButton /> }}` />`.
And by default it will use our components. So in `notification-internal` it would be something like this:
```
const { CloseButton: CloseButtonOverride } = overrides;

 // isValidElement from React
 // InternalCloseButton is Notification.CloseButton
const CloseButton = isValidElement(CloseButtonOverride) || InternalCloseButton;
```
It could solve the issue with (although rare) styling / customizing internals for some more complex components (lika Dialog).

What do you think? Cold be nice experiment, and later we can improve other components with it potentially.

src/notification/notification-internal.tsx
```diff
@@ -18,15 +18,27 @@ type NotificationProps = {
...
+  /* Custom close button override. In most cases, still use `Notification.CloseButton`. */
+  closeButton?: JSX.Element;
+} & NotificationBaseProps;
+
+type NotificationPersistent = {
+  closable?: never;
+  closeButton?: never;
+} & NotificationBaseProps;
+
+type NotificationProps = NotificationClosable | NotificationPersistent;
```
blurbyte, line 40: Don't think we have to be that strict here. If someone will make `closable` conditional it will make their life harder overall.

blurbyte (changes requested):
Well done. Pretty much what we need here.

Maybe we should make it a bit more generic and introduce simple `overrides` API (basically collection of predefined slots). Such consistent pattern of internal components can be used to improve other complex components. During next steering group we can for example ask ppl for their preferred candidates for adding it.

src/notification/__tests__/notification.spec.tsx
```diff
@@ -52,6 +52,25 @@ describe('Notification', () => {
     expect(button).toBeInTheDocument();
   });
 
+  it('renders custom close button', async () => {
```
blurbyte, line 55: Maybe would be worth testing if `onClose` and custom button `onClick` are triggered correctly?

blurbyte (changes requested):

src/notification/notification-internal.tsx
ixTec: As discussed, leaving this as future work for when we overhaul typings and apis for components

src/notification/notification-internal.tsx
ixTec: `closeButton` shouldn't be provided if the the Notification isn't `closable`. Should be very easy to handle as long as they wrap both `closable` and `closeButton` in whatever conditional they use

blurbyte: I just think technically there is no harm if someone whats to pass custom button which is not even visible, most likely for future use e.g. to conditionally switch between closable. But I'm also fine with leaving this more advanced typings, it just could be a bit cumbersome from DX perspective.

src/notification-card/notification-card-internal.tsx
```diff
@@ -29,6 +31,25 @@ function NotificationCardInternal(
 ) {
   const isIconCentered = closable && isContentCentered;
 
+  // Maintain the original onClose behavior, if a custom one is provided
+  function handleClick() {
```
blurbyte, line 35: I would refactor whole close button handling to something a bit more cleaner, e.g.
```
function renderCloseButton() {
    function handleClick() {
      closeButtonOverride?.props.onClick?.();
      onClose?.();
    }

    if(closable) {
      return closeButtonOverride ? cloneElement(closeButtonOverride, {
        inverted,
        onClick: handleClick
      }) : <CloseButton inverted={inverted} onClick={onClose} />
    }

    return null;
  }
```
and in component itself just 
```
<section>
  ...
  {renderCloseButton()}
</section>
```

src/notification-card/notification-card.tsx
```diff
@@ -1,7 +1,12 @@
 import { forwardRef } from 'react';
 
 import NotificationCardInternal from './notification-card-internal';
+import CloseButton from './close-button';
 
 const NotificationCard = forwardRef(NotificationCardInternal);
 
-export default NotificationCard;
+const NotificationCardNamespace = Object.assign(NotificationCard, {
+  CloseButton,
```
blurbyte, line 9: Should we mention this subcomponent in docs (probably not, but not sure if omitting it was intentional)?

src/notification/__tests__/notification.spec.tsx
```diff
@@ -52,6 +52,49 @@ describe('Notification', () => {
     expect(button).toBeInTheDocument();
   });
 
+  it('renders custom close button with onClose and onClick handlers working', async () => {
```
blurbyte, line 55: Maybe 'when custom close button is provided, appropriate onClose and onClick handlers are fired'?

src/notification/notification-internal.tsx
```diff
@@ -18,15 +18,27 @@ type NotificationProps = {
...
   onClose?: () => void;
   /* Custom Notification action. Use `Notification.ActionButton` subcomponent. Must be a single element. */
   action?: JSX.Element;
 } & React.HTMLAttributes<HTMLDivElement>;
 
+type NotificationClosable = {
+  /* Shows a close button, and the Notification can be dismissed by a user. */
+  /* @default false */
+  closable: boolean;
+  /* Custom close button override. In most cases, still use `Notification.CloseButton`. */
```
blurbyte, line 31: Maybe 'Custom close button component. In general use `Notification.CloseButton` subcomponent.'?

doc-site/examples/notification/custom-close.tsx
```diff
@@ -0,0 +1,15 @@
+import { Notification } from '@datacamp/waffles/notification';
+
+function Example() {
+  return (
+    <Notification
+      title="This course has been updated recently"
+      closable
+      closeButtonOverride={
+        <Notification.CloseButton data-trackid="custom track id" />
```
blurbyte, line 9: Data attrs should be in general kebab cases, so `custom-track-id`.

src/notification/__tests__/notification.spec.tsx
```diff
@@ -52,6 +52,49 @@ describe('Notification', () => {
...
+    const handleClick = jest.fn();
+    const handleClose = jest.fn();
+    const { getByTestId, queryByText } = render(
+      <Notification
+        title="Taylor Swift tour announced"
+        closable
+        onClose={handleClose}
+        closeButtonOverride={
+          <Notification.CloseButton
+            data-testid="Custom close button"
```
blurbyte, line 67: Should be `custom-close-button`.

blurbyte (changes requested):
We are really close to merging. Have just a few minor remarks. 
I would maybe consider changing `closeButtonOverride` to just `closeOverride`. It just gives me some strong jQuery vibes for whatever reason :). But be free to leave it more explicit, not a big deal.

src/notification/card.tsx
```diff
@@ -24,7 +24,8 @@ type CardProps = {
   variant: NonNullable<React.ComponentProps<typeof Notification>['variant']>;
   inverted: boolean;
   closable: boolean;
-  onClose?: () => void;
+  onClose: () => void;
```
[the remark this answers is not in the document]
ixTec: It will always be provided, as `notification-internal` has the `handleClose` function passed through

src/notification-card/notification-card.tsx
ixTec: Unintentional, so I've added it to docs! Best to have it documented, just like dialog close button

ixTec:
Leaving `closeButtonOverride` as is, because I think `closeOverride` is a little confusing (could imply it overrides close behaviour etc)

blurbyte (approved):
Let's merge 🦀 .

https://media.giphy.com/media/LcfBYS8BKhCvK/giphy.gif

PR #124, opened 2022-11-22
[WF-705] Rename Toast/Notification `default` variant to `info`

ixTec:
# Rename Toast/Notification `default` variant to `info`

### 🚨 Breaking Changes

- Rename Notification `default` variant to `info`
- Rename Toast `default` variant to `info`

### 🔧 Minor Changes

- Notification `title` is now of type `React.ReactNode` (was `string`)
    - We could add an example for a custom title to the docs, but I think this is a pattern we don't want to actively promote


blurbyte (approved):
Looks good 👍 . I will create a story to update it in Figma too. I agree we shouldn't add extra example for notification title, since that's something we don't want to promote overall.

PR #126, opened 2022-11-23
[WF-674, WF-686, WF-688, WF-700] Update Dialog internals, improve accessibility and add decorative header variant

ixTec:
# Update Dialog internals, improve accessibility and add decorative header variant

- [WF-674](https://datacamp.atlassian.net/browse/WF-674)
- [WF-686](https://datacamp.atlassian.net/browse/WF-686)
- [WF-688](https://datacamp.atlassian.net/browse/WF-688)
- [WF-700](https://datacamp.atlassian.net/browse/WF-700)

### 🚨 Breaking Changes

- `Dialog`, `AlertDialog` and `Drawer` now have their `id`, `aria-describedby` and `aria-labelledby` set automatically internally

### 🚀 Major Changes

- `Dialog.Button` now supports polymorphic typing via the `as` prop
- `Drawer `now relies more heavily on `Dialog` internally
- `AlertDialog` now directly uses `Dialog.Footer` and `Dialog.Body` internally
- `Dialog`, `AlertDialog` and `Drawer` now support `closeButtonOverride` prop, for custom close button 

### 🔧 Minor Changes

- `Dialog` content and buttons can now be center aligned via `alignCenter` prop
- Removed `border-top` from `Dialog.Footer` and `Drawer.Footer`
- `Overlay` now uses `floating-ui` overlay as a base
- Extracted Icon logic from `Notification` into separate `NotificationIcon` component
- Fixed issue with icon colour for `upgrade` variant `Notification`

### ❓ Misc Changes

- Removed `react-focus-on` dependency. Functionality has been replaced/refactored using `floating-ui`
- Updated `Drawer` docs to reflect `Dialog` (removed most prop tables)
- Fixed casing within buttons of several examples

## Screenshot(s):

Screenshot 2022-11-29 at 16 00 24
Screenshot 2022-11-29 at 16 03 03
Screenshot 2022-11-29 at 16 00 32
Screenshot 2022-11-29 at 16 01 43
Screenshot 2022-11-29 at 16 01 50
Screenshot 2022-11-29 at 16 01 57
Screenshot 2022-11-29 at 16 02 03
Screenshot 2022-11-29 at 16 04 01


blurbyte (changes requested):
Looks really nice so far 🎉 ! Addressed some Visual and Examples issues first. 

## Visual

1. All buttons for `Dialog` and `Drawer` should be right aligned with main CTA as the rightmost one (when not `alignCenter`), with `spacing.tokens.small` (8px) gap between them
2. Bottom padding of `Dialog.Footer` and `Drawer.Footer` should be increased to `spacing.tokens.large` (24px)
3. I think `CloseButton` in regular and decorative headers should be in the same position - I'd suggest `top: 6px` and `right: 6px`.
4. Decorative header height is incorrect - should be `44px` of dark bgr and `4px` of color decor. Although here I would try `CloseButton` placement from point 3. and than adjust dark bgr height, so X is exactly in the middle.
5. Missing thin border around decorative header (top, left, right, 15% white)
6. Missing hover color for decorative header `CloseButton` - should 10% white
7. On _mobile_, layout of buttons in Dialog / Drawer / AlertDialog `Footer` should be as following:
For 2 buttons they should be next to each other but taking all the space (most likely `grid-template-columns: 1fr 1fr` should be enough):
Screenshot 2022-11-30 at 14 35 17
For 3+ buttons they should be stacked vertically (ignore incorrect spacing in the example, should always be 8px):
Screenshot 2022-11-30 at 14 39 03

## Examples

1. Please don't use 'Lorem ipsum' in examples, shouldn't be hard to find something in our products copy. In tests it's fine.
2. In 'Decorative header' example I would suggest using radios to set preferred variant as in Menu 'Placement':
Screenshot 2022-11-30 at 14 22 40

src/notification-icon/notification-icon.tsx
```diff
@@ -0,0 +1,49 @@
...
+  CrossCircleInverted,
+  InfoCircleInverted,
+  RocketInverted,
+} from '../icon';
+
+import { iconBackgroundStyle, iconStyle, iconWrapperStyle } from './styles';
+
+type NotificationIconProps = {
+  variant?: 'info' | 'success' | 'warning' | 'error' | 'upgrade';
+  size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
```
blurbyte, line 15: Do we need all these sizes here? I think currently we are using only `medium` and `large`. Would make `backgroundStyle` a bit more straightforward.

src/notification-icon/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const invertedIconVariantMap = {
+  ...regularIconVariantMap,
+  upgrade: {
+    ...regularIconVariantMap.upgrade,
+    color: tokens.colors.purpleLight,
+  },
+};
+
+export function iconWrapperStyle() {
```
blurbyte, line 42: I think since it's separate component now (so we are in 'icon' context) we can easily call it just `wrapperStyle`.

src/notification-icon/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type IconBackgroundStyleOptions = {
+  variant: NonNullable<
+    React.ComponentProps<typeof NotificationIcon>['variant']
+  >;
+  size: NonNullable<React.ComponentProps<typeof NotificationIcon>['size']>;
+  inverted: boolean;
+};
+
+export function iconBackgroundStyle({
```
blurbyte, line 86: Similar to wrapper, we can call it `backgroundStyle`.

src/notification-icon/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+  variant,
+  size,
+  inverted,
+}: IconBackgroundStyleOptions) {
+  // Calculate icon-size relative background sizing
+  const radius =
```
blurbyte, line 92: Imho I would just handle 2 cases explicitly, the ones which are currently in use: medium and large. Or if we want to keep all these sizes probably it would be more straightforward to just declare in our usual map:
```
const radiusSizeMap = {
  xsmall: 4,
  small: 6,
  medium: 6
  ...,
  xlarge: 8
}
```

src/overlay/overlay.tsx
```diff
@@ -7,7 +8,13 @@ type OverlayProps = {
 } & React.HTMLAttributes<HTMLDivElement>;
 
 function Overlay({ isVisible, ...restProps }: OverlayProps) {
-  return <div {...restProps} css={overlayStyle({ isVisible })} />;
+  return (
+    <FloatingOverlay
```
blurbyte, line 12: Really nice, simplified styling greatly 👍 .

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,103 @@
...
+  closeButtonOverride?: JSX.Element;
+  /* Whether to center align the Dialog content and buttons. */
+  /* @default false */
+  alignCenter?: boolean;
   /* Content of the Dialog. In general, Dialog's own subcomponents should be used: `Dialog.Header`, `Dialog.Body`, and `Dialog.Footer`. */
   children: React.ReactNode;
+  /* [skip docs] */
+  role?: 'dialog' | 'alertdialog';
+  /* [skip docs] */
+  idPrefix?: string;
```
blurbyte, line 35: Not really sure if we need it. Don't think it really matters if it will be different for `AlertDialog` and `Drawer`. Should be fine with `modal` for all cases.

blurbyte (changes requested):

src/dialog/close-button.tsx
```diff
@@ -5,19 +5,21 @@ import { Button } from '../button';
...
+  hasDecorativeHeader?: boolean;
+} & Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'children'>;
 
-function CloseButton(props: CloseButtonProps) {
+function CloseButton({
+  hasDecorativeHeader = false,
+  ...restProps
+}: CloseButtonProps) {
   return (
     <Button
```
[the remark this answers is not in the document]
ixTec: Completely overlooked this 🤦‍♂️

blurbyte (changes requested):
Some general remarks about tests and code structure.

1. Really love that it was possible to keep the same API for footer buttons 👏 
2. Really clean appraoch with a context overall
3. Not really sure if we need ID prefix and different IDs for heading and body
4. `Drawer` is rather broken. There was a reason it had these nested wrappers. 
Anyway it should look like this on mobile, aside some stylistic changes to footer:
Screenshot 2022-12-01 at 15 51 02
But looks like this:
Screenshot 2022-12-01 at 15 50 56
5. Header variants examples and e2e could be improved slightly

src/notification-icon/notification-icon.tsx
ixTec: Fair point, I can tidy this up along with the idea of a map for the values instead

src/dialog/footer.tsx
```diff
@@ -1,16 +1,36 @@
...
 } & React.HTMLAttributes<HTMLDivElement>;
 
 function Footer({ children, ...restProps }: FooterProps) {
+  const { alignCenter, autoFocusRef } = useDialog();
+
+  // Add a reference to the button with autoFocus (if present), so that it can be focused on open
+  function renderChildren() {
+    return Children.map(children, (child) => {
+      if (isValidElement(child)) {
+        return child.type === Button && child.props.autoFocus
```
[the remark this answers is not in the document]
ixTec: This is a fair point, although I believe we shouldn't be trying to focus non-button content either...

The `autoFocus` prop is a custom prop for our `Button` and isn't the native behaviour (which only autofocuses first instance on page load), so I think it's a good idea to restrict the attempted behaviour to our Buttons still. What do you think?

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,103 @@
...
+  role = 'dialog',
+  idPrefix = 'dialog',
   ...restProps
 }: DialogProps) {
   const isAnimating = useAnimateTransition(isOpen, 300);
+  const id = `${idPrefix}-${useId()}`;
+  const autoFocusRef = createRef<HTMLButtonElement>();
+
+  // Determine if `children` contains `Header` and/or `Body` components
+  function childConfig() {
```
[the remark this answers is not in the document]
ixTec: Looking at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby, it seems aria-describedby should be provided unless the content does not have a descriptive purpose. In your example, where forms etc are provided then a consumer should provide `aria-label` on the `Dialog` instead. I'll add support to not add `aria-describedby` if `aria-label` is provided. 

As for the actual function, we could refactor into a hook here, as long as it reruns if `children` changes. I'll also make this improvement

src/dialog/dialog-internal.tsx
ixTec: I think `modal` is too generic for what we're doing, and it doesn't hurt to be more specific with our ids here, especially where some have styling overwritten etc

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,103 @@
...
-  role = 'dialog',
+  closeButtonOverride,
+  alignCenter = false,
   children,
+  role = 'dialog',
+  idPrefix = 'dialog',
   ...restProps
 }: DialogProps) {
   const isAnimating = useAnimateTransition(isOpen, 300);
+  const id = `${idPrefix}-${useId()}`;
```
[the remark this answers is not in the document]
ixTec: Raised https://datacamp.atlassian.net/browse/WF-711 to handle future `Modal` component refactor

src/dialog/dialog-internal.tsx
ixTec: Raised https://datacamp.atlassian.net/browse/WF-711 as discussed

doc-site/examples/dialog/decorative-header.tsx
```diff
@@ -0,0 +1,197 @@
+import { useState } from 'react';
+import { css } from '@emotion/react';
+import { tokens } from '@datacamp/waffles/tokens';
+import { Paragraph } from '@datacamp/waffles/paragraph';
+import { Dialog } from '@datacamp/waffles/dialog';
+import { Button } from '@datacamp/waffles/button';
+
+function Example() {
+  const [isOpen, setIsOpen] = useState([false, false, false, false, false]);
```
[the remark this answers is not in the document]
ixTec: Great suggestion!

src/drawer/drawer-internal.tsx
```diff
@@ -1,17 +1,16 @@
-import { FocusOn } from 'react-focus-on';
 import React from 'react';
 
-import { Portal } from '../portal';
-import { Overlay } from '../overlay';
-import { useAnimateTransition } from '../hooks';
+import { Dialog } from '../dialog';
 
-import Panel from './panel';
```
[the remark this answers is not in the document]
ixTec: The same structure is used within Dialog (overlay and panel separate), so shouldn't actually be any different, but I don't have a device to test for sure unfortunately.

I fixed Drawer responsive behaviour overall, as it was actually still broken before and not adhering to max-width for below medium

src/dialog/footer.tsx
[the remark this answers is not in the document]
blurbyte: It's fine :)

src/notification-card/styles.ts
```diff
@@ -69,117 +69,19 @@ export function decorStyle({ variant, inverted }: DecorStyleStyleOptions) {
...
-    position: relative;
-    z-index: ${tokens.zIndex.default};
-    width: ${tokens.sizing.small};
-    height: ${tokens.sizing.small};
-    flex-shrink: 0;
-
     ${isIconCentered &&
     css`
-      margin-left: auto;
+      /* margin-left: auto; */
```
blurbyte, line 80: It's commented out, so I assume we don't need it at all.

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,96 @@
...
+  'aria-label': ariaLabel,
+  role = 'dialog',
+  idPrefix = 'dialog',
   ...restProps
 }: DialogProps) {
   const isAnimating = useAnimateTransition(isOpen, 300);
+  const autoFocusRef = createRef<HTMLButtonElement>();
+  const id = `${idPrefix}-${useId()}`;
+  const [headerId, setHeaderId] = useState<string>();
+  const [bodyId, setBodyId] = useState<string>();
```
blurbyte, line 55: I'm not convinced we need `bodyId` at all (same for `idPrefix`). Anyway I think we will want 1 more PR for modals anyway to put all shared stuff in `Modal` internal component. Also most likely we want to expose `useDialog` hook, which has some other implications.
So for now we can leave it as it is, and discuss these 2 things tomorrow.

blurbyte (changes requested):
We are pretty close to merging it. Got some ideas around `idPrefix` and `bodyId`, but I guess it could be tackled in next PR.

doc-site/pages/components/dialog.mdx
```diff
@@ -8,24 +8,47 @@ import Imports from '../../components/imports';
...
 
+`Dialog` must have a `Dialog.Header` child component provided in order to meet modal accessibility requirements via `aria-describedby`.
+
 It is recommended to have the initial focus set on the most relevant element&mdash;in most cases confirmation button. To do so, pass `autoFocus` flag to `Dialog.Button` subcomponent used in the dialog's footer.
 If no initial focus is set explicitly close button will be focused by default.
 
-To properly label `Dialog` and provide context for users using screen readers, associate both header and dialog by unique identifier by adding `id` attribute to `Dialog.Header` and `aria-labelledby` to `Dialog` itself. If the design doesn't include a visible label provide `aria-label` to `Dialog` instead.
-
-Additionally, a description may be provided with `aria-describedby` associated with one of the `Dialog.Body` paragraph's `id`.
+If `Dialog.Body` contains non-descriptive content (e.g. form elements), then `aria-label` should be provided on the main `Dialog` element to provide context of the content.
```
blurbyte, line 25: I'd say `Notification` should be used here.

src/dialog/header.tsx
```diff
@@ -1,19 +1,64 @@
...
-function Header({ children, ...restProps }: HeaderProps) {
+type HeaderDecorativeProps = {
+  /* Mode of the dialog. */
+  /* @default plain */
+  mode: 'plain' | 'decorative';
+  /* Variant of the dialog, only applicable when `mode="decorative"`. */
+  /* @default info */
+  variant?: 'info' | 'success' | 'warning' | 'error' | 'upgrade';
+  /* Custom icon override for the decorative header, only applicable when `mode="decorative"`. In general, use Waffles [Icon](/components/icon). */
+  customIconOverride?: JSX.Element;
```
blurbyte, line 22: I think calling `customIcon` should be enough.

src/dialog/header.tsx
```diff
@@ -1,19 +1,64 @@
...
-type HeaderProps = {
-  /* Title of the modal. */
+type HeaderBaseProps = {
+  /* Title of the dialog. */
   children: React.ReactNode;
 } & React.HTMLAttributes<HTMLHeadingElement>;
 
-function Header({ children, ...restProps }: HeaderProps) {
+type HeaderDecorativeProps = {
+  /* Mode of the dialog. */
```
blurbyte, line 15: Probably would be good idea to explain a bit more (with 1 extra sentence or sth) what it actually does.

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,105 @@
...
-  role = 'dialog',
+  closeButtonOverride,
+  alignCenter = false,
   children,
+  'aria-label': ariaLabel,
+  role = 'dialog',
   ...restProps
 }: DialogProps) {
   const isAnimating = useAnimateTransition(isOpen, 300);
+  const didMount = useRef(false);
```
blurbyte, line 52: Would put some new lines between some related items to 'group' them a bit.

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,105 @@
...
+  useRef,
+  useState,
+} from 'react';
 
 import { Portal } from '../portal';
 import { Overlay } from '../overlay';
-import { useAnimateTransition } from '../hooks';
+import { useAnimateTransition, useId } from '../hooks';
+import { logError } from '../helpers';
+import AlertHeader from '../alert-dialog/header';
```
blurbyte, line 14: Would add TODO here, since we usually don't import components that way. Probably will be solved nicely after refactoring shared stuff into `Modal`.

blurbyte (changes requested):
Just a few very minor things to update.

src/dialog/debug-messages.ts
```diff
@@ -0,0 +1,5 @@
+// Constants to be used as content for the associated component's `console.error` or `console.warn`.
+
+export const MESSAGES = {
+  MISSING_HEADER: 'Dialog requires a `Dialog.Header` child component.',
```
blurbyte, line 4: We can add a bit of reasoning here, e.g. `For accessibility reasons...`

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,105 @@
...
+      if (isValidElement(child)) {
+        if (child.type === Header || child.type === AlertHeader) {
+          setHeaderId(`${id}-header`);
+          setHasDecorativeHeader(child.props.mode === 'decorative');
+        } else if (child.type === Body || child.type === AlertBody) {
+          !ariaLabel && setBodyId(`${id}-body`);
+        }
+      }
+    });
+  }, [ariaLabel, children, headerId, id]);
```
blurbyte, line 72: Don't think `headerId` is needed here.

src/dialog/dialog-internal.tsx
```diff
@@ -1,56 +1,105 @@
...
   children,
+  'aria-label': ariaLabel,
+  role = 'dialog',
   ...restProps
 }: DialogProps) {
   const isAnimating = useAnimateTransition(isOpen, 300);
+  const didMount = useRef(false);
+  const autoFocusRef = createRef<HTMLButtonElement>();
+  const id = `modal-${useId()}`;
+  const [headerId, setHeaderId] = useState<string>();
```
blurbyte, line 57: I think setting `null` as initial value for both `headerId` and `bodyId` makes it a bit clearer it _might_ not exist.

src/dialog/button-internal.tsx
```diff
@@ -1,46 +1,59 @@
...
+function ButtonInternal<T extends React.ElementType = 'button'>(
+  {
+    as,
+    variant = 'primary',
+    size = 'medium',
+    isLoading = false,
+    disableTitleCase = false,
+    iconLeft,
+    iconRight,
+    autoFocus,
```
blurbyte, line 33: I think we can easily omit all of these here and leave only `as`, `children`, and `...restProps`.

blurbyte (changes requested):

src/dialog/debug-messages.ts
ixTec: Leaving as is, as discussed, as the documentation covers a more detailed explanation and these errors should be kept simple for the most part

blurbyte (approved):
Let's merge 🚀 

cocktail party

PR #130, opened 2022-12-06
[WF-713] Link underline fix for Safari

blurbyte:
# Fixed link underline animation for Safari

### 🔧 Minor Changes

- Fixed `Link` underline animation in Safari, by changing the timing of the transition, so it's hiding properly; before animation was ending at around 20% opacity
- Updated `Link` underline thickness so it's uniform for all sizes (it was barely noticeable for small size)

## Screenshot(s):

Underline thickness adjustments:

After:
uniform-underline

Before:
auto-underline


ixTec (approved):
Looks good to me 👍

PR #135, opened 2022-12-28
[WF-736] Downgrade `floating-ui` dependency

ixTec:
# Fix for focus issues as a result of upgrading `floating-ui-` dependency

## [WF-736](https://datacamp.atlassian.net/browse/WF-736)

Requesting review from @lensbart during team absences, as he raised the original issue with Waffles v1.17.0

### 🔧 Minor Changes

- Downgraded `floating-ui` dependency back to v`0.10.3`
    - Further changes required to handle breaking changes in `floating-ui` when upgrading the dependency next

### ❓ Misc Changes

- Re-enabled release in pipeline


lensbart (approved):
L👍👍TM

PR #137, opened 2022-12-29
[WF-738] Fix type of `title` prop for both `EmptyState` and `Notification`

lensbart:
# Fix type of `title` prop for both `EmptyState` and `Notification`

### 🔧 Minor Changes

- The type of `EmptyStateBaseProps` had a bug because of `title` being a property of `React.HTMLAttributes<HTMLDivElement>`. This fix will enable supplying a `title` of type `ReactNode` to `EmptyState`, as intended. This fix also includes a restructuring of most props, to prevent a similar bug from being introduced in the future.

## Screenshot(s):

Not applicable

[WF-738]: https://datacamp.atlassian.net/browse/WF-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WF-738]: https://datacamp.atlassian.net/browse/WF-738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ixTec (changes requested):
So overall using `Omit<T, keyof ComponentProps>` is quite an elegant safety feature for this sort of issue, however unfortunately the use of `keyof` is not currently supported in the dependency we use for generating our component docs prop tables. If you try to build and run the doc-site locally, you'll see we get an error for this (https://github.com/atlassian/extract-react-types/issues/149). This is why we have to omit a list of explicit props.

For this PR, we'll have to do the same for the `EmptyState` component as well.

I'm also not a fan of the separate type declaration for this and so would prefer we keep the `Omit` as part of the base type, e.g. as we had it for some components, so for example:
This fits more closely with how we structure our types across components, reserving additional type declarations for complex/polymorphic types.

lensbart:
@ixTec Thanks for the detailed description. I’ve amended the change to now include just a manual fix, for both `EmptyState` and `Notification`, which had the same problem.

ixTec (approved):
LGTM 🚀

PR #139, opened 2023-01-03
[WF-662] fix `setTitleCase` to not force-lowercase the “C” in “DataCamp” or force-uppercase the “i” in “iPhone”

lensbart:
# Fix `setTitleCase` to not force-lowercase the “C” in “DataCamp” or force-uppercase the “i” in “iPhone”

### 🔧 Minor Changes

- This PR changes the capitalisation behaviour for button text. With these changes, already-capitalised words (e.g. `iPhone`, `SQL`, `DataCamp`, `GitHub`, etc.) will remain intact, provided that they don’t appear in the list of stop words. Previously, they were changed to respectively `Iphone`, `Sql`, `Datacamp` and `Github` and the `disableTitleCase` prop/manual capitalisation was required. Now, it should work out-of-the-box.

[WF-662](https://datacamp.atlassian.net/browse/WF-662) contains two items: the fix made in this PR, and a suggestion to decrease the amount of stop words so as to conform to an existing (or more widely used) style guide. Because of this suggestion, there was a need for further consideration, which (I’m assuming) caused the ticket to be moved to the backlog. Without a change in stop words, this PR is an unambiguous improvement to `setTitleCase`. (For context: at Workspace, we have 5 instances of `setTitleCase` that were introduced to offset this bug.)

I’ve created [WF-744](https://datacamp.atlassian.net/browse/WF-744) to track this second item.

[WF-662]: https://datacamp.atlassian.net/browse/WF-662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WF-662]: https://datacamp.atlassian.net/browse/WF-662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WF-662]: https://datacamp.atlassian.net/browse/WF-662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WF-662]: https://datacamp.atlassian.net/browse/WF-662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WF-744]: https://datacamp.atlassian.net/browse/WF-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/helpers/set-title-case.ts
```diff
@@ -38,6 +38,20 @@ export const TITLE_CASE_STOP_WORDS = [
   'of',
 ];
 
+function hasUppercase(text: string) {
```
ixTec, line 41: Let's add a single line comment here for what it's for

src/helpers/set-title-case.ts
```diff
@@ -38,6 +38,20 @@ export const TITLE_CASE_STOP_WORDS = [
...
+      letter !== letter.toLowerCase()
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
+
 /**
  * Convert string content into Title Case
```
ixTec, line 56: We likely want to extend this comment to reflect the change that mixed case words will be respected / left unchanged

ixTec (changes requested):
A nice solution for handling mixed case words. We may want to approach this differently (perhaps with configuration) in the future, as words "liKE tHIS" will be left as is (which would be undesired).  I don't think this is something we need to address now though.

Besides the few inline comments I've added, I think we should also update the documentation example for this (`doc-site/examples/helpers/set-title-case-basic-usage.tsx`), to include some mixed case words. Maybe some copy like "Download DataCamp for Mobile today. Available on iPhone and Android"

src/helpers/set-title-case.ts
lensbart: “Mixed case” would imply the presence of both lowercase and uppercase characters, whereas this function only reports `true` if it encounters the presence of uppercase characters.

I found the name of this function to be sufficiently self-explanatory, but happy to add a comment. Is the inclusion of a comment per function always desired? Then ideally this would be reflected in `CODING_STYLE.md`.

src/helpers/set-title-case.ts
```diff
@@ -38,6 +38,20 @@ export const TITLE_CASE_STOP_WORDS = [
   'of',
 ];
 
+function hasUppercase(text: string) {
+  for (let index = 0; index < text.length; index += 1) {
```
ixTec, line 42: `index += 1` can be shortened to `index++` notation

src/helpers/set-title-case.ts
ixTec: You're right, that's my bad. I think `containsUppercase` or `includesUppercase` would be more suitable here though.

For comments, it's just something we tend to do for our helper / hook utils - not across the rest of the repo

lensbart:
> words "liKE tHIS" will be left as is (which would be undesired)

@ixTec I think it will be impossible to police against this, and we should trust the developer here. (It is very unlikely that anyone would write something like this by mistake, or without good reason.)

ixTec (approved):

PR #140, opened 2023-01-04
[WF-742] No More Beige token updates

ixTec:
# Update greys, `navySubtleTextOnLight` and `blueDarkText` token colors for the no more beige work

### 🚨 Breaking Changes

- Updated colors for the following tokens:
    - `greyLight`: `#EFEFF5`
    - `grey`: `#E8E8EE`
    - `greyMedium`: `#E1E1E8`
    - `greyDark`: `#848492`
    - `blueDarkText`: `#0065D1`
    - `navySubtleTextOnLight`: `#5D6A77`

### 🚀 Major Changes

- Added the following new tokens:
    - `transparentNavy`: `rgba(5, 25, 45, 0.15)`
    - `transparentNavySubtle`: `rgba(5, 25, 45, 0.1)`
    - `transparentWhite`: `rgba(255, 255, 255, 0.15)`
    - `transparentWhiteSubtle`: `rgba(255, 255, 255, 0.1)`

[WF-742]: https://datacamp.atlassian.net/browse/WF-742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lensbart:
This is a case where `@datacamp/eslint-plugin-waffles` can shine, by adapting [this file](https://github.com/datacamp-engineering/collab-and-tooling/blob/490e2b354fa9aedc777b2793e32ab053017ebd0c/libs/config/eslint-plugin-waffles/src/rules/tokens/utils/getColorReplacement.ts#L175) so that hard-coded deprecated color values can be replaced by the token, so that the new color is used.

blurbyte (changes requested):
Looks good! As we introduced new toxic blue, I think we can push it a bit further and:
1. Update docs to reflect changes: https://waffles-next.datacamp.com/foundation/design-tokens
2. Update the background to be `greySubtle`, replace beige in docs components with greys
3. Update the color of red dot indicator (I think it was suppose to be `redDark`, but let's double check with Maria)
4. Update the HEX value of `navySubtleTextOnLight`
5. Introduce new utility transparent tokens for border and hover:
    - transparentNavy (15% opacity navy, update borders)
    - transparentNavySubtle (10% opacity navy, update hover)
    - transparentWhite (15% opacity white, update inverted borders)
    - transparentWhiteSubtle (10% opacity white, update inverted hover)

ixTec:
Let's raise additional tickets for this work and then we can prioritise it alongside this work. A lot has changed since I raised this PR 😅

blurbyte (approved):
Looks good! I would maybe omit 'For smaller UI elements, typically `greyLight` or `grey` should be used.' from design tokens guidelines. Make sense to put new tokens in new utility subgroup 👍 .

ixTec:
Te updated docs content came from Maria/Greg so I'll leave it as is and then we can always review at a later point if needed

PR #154, opened 2023-02-13
[WF-777] Tag Component

ixTec:
# Added new Tag component

### 🚀 Major Changes

- New `Tag` component available

### ❓ Misc Changes

- Updated `Badge` usage guidance to reflect new addition of Tag component
- Fixed minor issue with css string types in `Card` css

## Screenshot(s):
Screenshot 2023-02-13 at 16 03 43

[WF-777]: https://datacamp.atlassian.net/browse/WF-777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tag/__tests__/tag-closable.story.tsx
```diff
@@ -0,0 +1,85 @@
...
+  'red',
+  'orange',
+  'yellow',
+  'blue',
+  'purple',
+  'pink',
+  'grey',
+] as const;
+
+function Story() {
```
blurbyte, line 34: I would maybe improve this example slightly to be more 'realistic'. We can put variants as initial state and remove appropriate one when tag is closed. That way good practice of keeping UI as a function of the state will be presented.

src/tag/__tests__/tag.e2e.ts
```diff
@@ -0,0 +1,19 @@
+describe('Tag', () => {
```
blurbyte, line 1: I would add test for focus style of interactive / closable tag.

src/tag/close-button.tsx
```diff
@@ -0,0 +1,66 @@
...
+  isSmallClosable: boolean;
+} & Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'children'>;
+
+function CloseButton({
+  size,
+  isSmallClosable,
+  ...restProps
+}: CloseButtonProps) {
+  // Render custom close icon at correct size
+  function renderCloseIcon() {
```
blurbyte, line 29: For clarity I would put it as a separate `CloseIcon` component (in this folder).

src/tag/close-button.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { ScreenReaderOnly } from '../screen-reader-only';
+import { Button } from '../button';
+
+import Tag from './tag-internal';
+import { closeButtonStyle } from './styles';
+
+const sizeMap = {
+  small: {
+    width: '12',
```
blurbyte, line 11: Because it's square would make sense to call it size / sizing and put as width / height.

blurbyte (changes requested):
Looks really good overall. Really good docs, distinction between Tag and Badge should be pretty clear.

Visually I think the fact that when hovering over X whole Tag is highlighted feels a bit awkward to me (focus is fine). For small Tag of course there is no way around it since hitbox is so small but for medium maybe only close button should have this hover style?

Close X for both sizes feels a bit too small.

Was thinking a bit more about API. Imho there is no real practical distinction between `onClick` and `onClose`, besides the part of element which is firing this event. Maybe it will be better to merge it into 1 handler called for example `onInteraction`. That way I think we can change `mode` into simple `closable` flag (to keep it more in line with Notification). Ideally probably `closable` should be default but `nonClosable` might be a bit weird. Also should help simplifying typings a bit. Or we can leave mode as it is to make difference between use cases a bit more apparent.

For whatever reason mode is described as string in docs (might be irrelevant if we proceed with changes above):
Screenshot 2023-02-14 at 13 24 23

src/tag/tag-internal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  function renderIcon() {
+    return icon
+      ? cloneElement(icon, {
+          css: iconStyle,
+          // Handle small buttons having xsmall sized Icons by default, and medium having small
+          ...(!icon.props.size && {
+            size: size === 'small' ? 'xsmall' : 'small',
+          }),
+        })
+      : '';
```
blurbyte, line 96: Probably should be _null_ instead of empty string.

blurbyte (changes requested):
We are super close to finishing it, really love how it looks and behaves.

Noticed there is no hover color change for 'navy' variant (should be lighter). Also I think we are missing some guidelines on capitalization of tag's label.

ixTec:
Updated hover styling for navy variant. Not sure we can specify any guidelines for content capitalisation and it's very dependent on what the content is

blurbyte (approved):
Looks good, let's merge 🚀

PR #165, opened 2023-03-07
[WF-708] Blues, hover and active styles update

blurbyte:
# Blues, hover and active styles update

## Changes

Most changes propagated automatically, after tokens were updated.

Some things to consider (will have a sync with Maria about some of these topics):

- [x] I feel hover is still a bit too light and should be slightly darker (**A:** decided on the new color)
- [x] Dark mode hover feels a bit too soft (**A:** no changes)
- [x] Tried adjusting on hover color of the link / plain button, but it made it very not ergonomic to use for custom buttons (most notably close buttons and in `Pagination`) and complicated code a bit too much: it won't be accessible on some backgrounds, but I don't think it's worth adjusting it, as it's necessary to 'reset' this color with pretty ugly CSS for each custom close button (**A:** decided to not update color on hover)
- [ ] using `HOVER_COLOR_LIGHTNESS_SHIFT` with `0.06` const in a few places, might be good idea to extract it to a token, but not sure how usable it will be for external users

### 🚨 Breaking Changes

- update set of our blue colors to better match recently adjusted `blueDarkText`
- adjusted values of the transparent color tokens
- rename `transparentNavy*` tokens to `transparentGrey*`

### 🚀 Major Changes

- slightly updated `Button` and `Tag` hover colors
- update ALPA Loop asset assessment blue color
- update `Menu` hover and active state, and category divider
- update `Table` light mode hover row color

### ❓ Misc Changes

- Mention new transparent soft colors usage in docs

## Screenshot(s):
IdX

[WF-708]: https://datacamp.atlassian.net/browse/WF-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ixTec (commented):
Looks really good overall, haven't spotted any obvious issues from my testing / exploration.

Few minor comments:

- PR should be targeting `rc-major` as it contains breaking changes
- Do we want to replace the colour we use for `ROW_HOVER_COLOR` in `Table` ? We may be able to use these new hover colours
- Haven't really noticed until now, but we probably need a better way of representing the `transparentWhite*` tokens, as right now they all (obviously) appear as just white 😄 Unsure of the best way of showing the usage here tbh
Screenshot 2023-03-22 at 15 32 39

ixTec (approved):

PR #167, opened 2023-03-10
[WF-778] Multiselect

ixTec:
# Add new Multiselect component and some additional minor fixes/changes

### 🚀 Major Changes

- Add new `Multiselect` component
- (internal impact only) Extracted `Menu` logic into standalone internal `Dropdown`
    - new `Multiselect` and `Menu` now use `Dropdown` internally

### 🔧 Minor Changes

- Fixed padding in `Menu` / `Dropdown` internal component, for items with large decor but no description
- Fixed issue with `Tag` when being  `closable` and `disabled` would not actually disable the close button
- Fixed unintentional propagation of click events on `Tag`

### ❓ Misc Changes

- Added support to docs prop table helper for complex types
    - A secondary table will be generated below the prop table, for any complex types (currently only found in `Multiselect`)

## TODO
- Add missing doc examples
- Complete unit/spec tests

## Screenshot(s):
Screenshot 2023-03-10 at 13 40 25
Screenshot 2023-03-10 at 13 40 55
Screenshot 2023-03-10 at 13 42 24
Screenshot 2023-03-10 at 13 42 35
Screenshot 2023-03-10 at 13 43 24
Screenshot 2023-03-10 at 13 43 36


doc-site/examples/multiselect/advanced.tsx
```diff
@@ -0,0 +1,106 @@
+import { useState } from 'react';
```
blurbyte, line 1: Maybe wrap Multiselect with `FormField` in this advanced example. Also maybe mark 1 more item with `isActive` to introduce some variety?

doc-site/examples/multiselect/basic-usage.tsx
```diff
@@ -0,0 +1,65 @@
+import { useState } from 'react';
+import { Multiselect, MultiselectItem } from '@datacamp/waffles/multiselect';
+
+const initialData: MultiselectItem[] = [
+  {
+    id: '0',
+    label: 'Python',
+    isActive: true,
```
blurbyte, line 8: I would leave all items inactive in this case, so we will have examples with 0, 1 and 2 items active initially.

src/multiselect/multiselect-internal.tsx
```diff
@@ -0,0 +1,242 @@
...
+  /* Decorative element displayed to the left of the Item. Could be any [Avatar](/components/avatar) or [Icon](/components/icon) from Waffles (use default size) or a custom component. */
+  decorLeft?: JSX.Element;
+  /* Whether the Item is active. */
+  isActive?: boolean;
+  /* Use to configure the Tag when the Item is active. These do not effect the dropdown Item. */
+  tag?: MultiselectTag;
+};
+
+type MultiselectBaseProps = {
+  /* Content for the Multiselect. */
```
blurbyte, line 50: Maybe mention it's similar to regular Select options, and that these are the items shown in the dropdown?

blurbyte (changes requested):
Overall component looks and works really, really nice 👏 . Haven't encountered any issues with how it's working so far. Leaving some general remarks. If something is really hard to change / implement please ignore it.

## Behavior

- I think input should be cleared after item got selected, right now whatever you typed is still there (we discussed it briefly, after trying it myself I don't like this behavior)
Screenshot 2023-03-15 at 14 23 33

- When input is focused focus ring is not visible in Safari and Firefox

- When clicking clear button input is not cleared
Screenshot 2023-03-15 at 14 36 01


## Look

- Chevron deffo looks different than what we have in Select (looks smaller and color seems to be different):

Multiselect:
Screenshot 2023-03-15 at 14 17 40

Select:
Screenshot 2023-03-15 at 14 17 59


- When dropdown item is hovered text is bold (should be regular)
Screenshot 2023-03-15 at 14 37 38

- I think when going through items in dropdown they should be highlighted in similar way to what we have in Menu, aka rendered with active item style but just show regular focus ring (I know technically they are not focused, but from visual consistency standpoint it makes more sense):

Multiselect:
Screenshot 2023-03-15 at 14 40 19

Menu:
Screenshot 2023-03-15 at 14 40 58

- Not a big fan of big left paddigns next to tag
Screenshot 2023-03-15 at 14 42 00

I think it looked much better in the earlier proposals (we should ask Maria for her opinion):
Screenshot 2023-03-15 at 14 43 00

Currently:

Select:
Screenshot 2023-03-15 at 14 17 59

- We should update `Input` example to use similar clear button as Multiselect:

Current input example:
Screenshot 2023-03-15 at 14 18 16

# A11y

- Well done! Overall works really well 🎉 
- I think simple 'Close tag' might be not enough as a label for Tag's close button: if wmultiple Tags it's announced in exactly the same way. Maybe we should announce whole label here 'Close Python tag' for example? Or in context of Multiselect I think it should be 'Remove python tag' or something similar. Actually for regular Tag it probably makes more sense to say 'Remove tag' instead of 'Close tag' too.

blurbyte (changes requested):
Really nice improvements 👍 .

Noticed pretty serious issue (which I think previously wasn't present):
- When removing 1 Tag it removes clicked one and the Tag following it, and the second item is completely gone (from dropdown): 

Initially:
Screenshot 2023-03-21 at 09 45 19

After SQL tag is removed:
Screenshot 2023-03-21 at 09 45 35

Really like how the Code Link looks like, but not a big fan of the hover style of the code link- underline should be enough:
Screenshot 2023-03-21 at 09 49 27

For medium and small Multiselects looks like top spacing is getting smaller when it has multiple lines of content.
Screenshot 2023-03-21 at 09 52 23

I think for small size gap between Tags should be slightly smaller.

Tag's close button is now really nicely announced 👍 .

I think after closing a Tag input should be focused back (currently looks like whole doc got focused and you have to tab again to focus Multiselect manually). If it's hard to implement skip it.

src/multiselect/tags.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+
+import { Multiselect } from './index';
+
+import { Tag } from '../tag';
+
+import { MultiselectItem } from './multiselect-internal';
```
blurbyte, line 7: If it's just a type it's better to explicitly import it as `import type { MultiselectItem } from ...`

src/multiselect/__tests__/multiselect-advanced.story.tsx
```diff
@@ -0,0 +1,130 @@
+import { useState } from 'react';
+import { css } from '@emotion/react';
+
+import { Multiselect, MultiselectItem } from '../index';
```
blurbyte, line 4: As `MultiselectItem` is type it's better to import it as `import type { MultiselectItem } from ...`

_Applies to other stories / examples too_

blurbyte (changes requested):

blurbyte (changes requested):
I think you wanted to expand tests a bit anyway, but to double check, we are missing a few e2es:
- starting from at lest 3 tags, test tags and dropdown menu items after tag is picked from dropdown
- starting from at least 3 tags, test tags and dropdown menu items after tag (some other tag than the last one) is removed
- clearing all tags with clear all button
- clearing input content with clear all button
- would make some simple test how it looks like on various device sizes (to see how multiline looks like)

doc-site/examples/multiselect/advanced.tsx
```diff
@@ -0,0 +1,141 @@
...
+          ? {
+              ...item,
+              isActive: !item.isActive,
+            }
+          : item;
+      }),
+    );
+  }
+
+  function handleClear() {
```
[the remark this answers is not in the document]
ixTec: Where `initialData` has some active items to begin with, setting the data back to `initialData` on clear, would not actually clear the active items.

I also think it would set a bad example to show setting the data back to the original value anyway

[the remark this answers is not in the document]
blurbyte: Makes sense 👍

blurbyte (approved):
Looks great 🥇.

Just 1 minor thing: looks like clear all button is misaligned slightly.
Screenshot 2023-03-22 at 17 45 38

celebration

PR #177, opened 2023-04-17
[WF-833] Minor Multiselect fixes/adjustments

ixTec:
# Fixes dropdown zindex, multiselect height and docs

### 🔧 Minor Changes

- Adjust `z-index` for dropdown (`Menu`/`Multiselect`) when inside a Waffles `Portal`-based component (e.g. `Dialog`).
-  Restrict height of `Multiselect` dropdown to prevent endless growth.
    - Overflow is automatic to allow for scrolling

### ❓ Misc Changes

- Add `onInput` prop for `Multiselect` to docs

## Screenshot(s):

[WF-833]: https://datacamp.atlassian.net/browse/WF-833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/dropdown/dropdown-internal.tsx
```diff
@@ -109,6 +111,17 @@ function DropdownInternal({
...
+  useEffect(() => {
+    // Determine whether the dropdown trigger is inside a Waffles Portal component (with default id)
+    if (isOpen && refs.domReference && !isParentModal) {
+      const modalParent = refs.domReference.current?.closest(
+        '#waffles-portal-root',
+      );
+
+      setIsParentModal(!!modalParent);
+    }
+  }, [isOpen, refs.domReference, isParentModal]);
```
lensbart, line 123: I think it would be more elegant to use an adapted (i.e., non-throwing) version of `useDialog`, or if that’s too big of a change, to use a custom hook like:

```ts
// src/dialog/dialog-context.tsx
export const useHasParentDialog = () => useContext(DialogContext) != null;
```

If pedantry is a forgivable offence, I would also prefer `has` to `is`, and `Dialog` to `Modal` (for consistency, although I prefer “modal” as a component name):

```ts
// src/dropdown/dropdown-internal.tsx
const hasParentDialog = useHasParentDialog()
```

lensbart: This would:

* be simpler
* avoid an implicit dependency on `#waffles-portal-root` id
* avoid an unnecessary rerender due to `useEffect`

src/dropdown/styles.ts
```diff
@@ -21,9 +21,17 @@ type DropdownStyleOptions = {
   x: number | null;
   y: number | null;
   inverted: boolean;
+  limitHeight: boolean;
```
lensbart, line 24:
```suggestion
  shouldLimitHeight: boolean;
```
makes it more clear that the value is a boolean, rather than a pixel amount

src/dropdown/styles.ts
```diff
@@ -21,9 +21,17 @@ type DropdownStyleOptions = {
...
+  isParentModal: boolean;
 };
 
-export function dropdownStyle({ x, y, inverted }: DropdownStyleOptions) {
+export function dropdownStyle({
+  x,
+  y,
+  inverted,
+  limitHeight,
+  isParentModal,
```
lensbart, line 33:
```suggestion
  hasParentDialog,
```

src/multiselect/multiselect-internal.tsx
```diff
@@ -60,6 +60,8 @@ type MultiselectBaseProps = {
   inverted?: boolean;
   /* Handler called when a dropdown Item is selected, or an active Tag is dismissed. Use to update the `isActive` prop for Items in `options`. */
   onChange: (id: string) => void;
+  /* Handler called a user types into the input of the Multiselect. Can be used to apply custom filtering behavior. */
```
lensbart, line 63: If you use [JSDoc syntax](https://jsdoc.app/about-getting-started.html), VSCode IntelliSense will pick it up and show it in a tooltip.

```suggestion
  /** Handler called a user types into the input of the Multiselect. Can be used to apply custom filtering behavior. */
```

lensbart, line 63: If you use [JSDoc syntax](https://jsdoc.app/about-getting-started.html), VSCode IntelliSense will pick it up and show it in a tooltip.

```suggestion
  /** Handler called a user types into the input of the Multiselect. Can be used to apply custom filtering behavior. */
```

lensbart (approved):
😎👍

src/dropdown/styles.ts
ixTec: Leaving as Modal, because it's used to refer to modal-like components (Alert Dialog, Dialog and Drawer) - which will be refactored to use an "modal" component based on Portal

src/dropdown/dropdown-internal.tsx
ixTec: I think this is a great suggestion. I've implemented this for now using the `*Dialog` suffix, but in the future I'd like to introduce a (internal, but may be shared publicly later on) `Modal` component that would be a better location for this hook (and then Dialog, Drawer etc build from this)

src/dropdown/styles.ts
```diff
@@ -3,10 +3,10 @@ import { css } from '@emotion/react';
 import { tokens } from '../tokens';
 
 export const DROPDOWN_MIN_WIDTH = 260;
+const DROPDOWN_MAX_HEIGHT = 268;
```
lensbart, line 6:
```suggestion
const VERTICAL_PADDING = tokens.spacing.small // use this variable on line 11
const MAX_NUMBER_OF_DROPDOWN_ITEMS = 7
/** In pixels, determined visually */
const DROPDOWN_ITEM_HEIGHT = 36 // assuming this value can’t be deduced from CSS properties
const maxDropdownHeight = MAX_NUMBER_OF_DROPDOWN_ITEMS * DROPDOWN_ITEM_HEIGHT + (2 * parseInt(VERTICAL_PADDING, 10))
```

lensbart (approved):

src/multiselect/multiselect-internal.tsx
lensbart: https://datacamp.atlassian.net/browse/WF-834

src/dropdown/styles.ts
ixTec: Added comment explaining use of it

PR #197, opened 2023-06-30
fix: add font family to modal style

Rahien:
# Insert Description Here

This is done so it renders text properly, even when not using a paragraph wrapper

### ❓ Misc Changes

Fix modal body to set font family, so text is styled properly when not using a paragraph to wrap the contents of a modal.

## Screenshot(s):
before:
image

after:
image

compared to with paragraph:
image

[WF-892]: https://datacamp.atlassian.net/browse/WF-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Rahien:
fixed in https://github.com/datacamp/waffles/pull/196

PR #220, opened 2023-09-01
fix(button): don't crash on multiple consecutive spaces in button text

BerndSchrooten:
An entire app can crash if text in a button contains multiple consecutive spaces. Workspace was crashing because we constructed a full name from a first name and last name that had trailing spaces. Ideally, this input is sanitized when the user enters his/her name in the platform as well so I'll flag that to responsible team as well.

### 🔧 Minor Changes

- Don't crash on multiple consecutive spaces in button text

## Screenshot(s):

[CT-8342]: https://datacamp.atlassian.net/browse/CT-8342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lensbart (approved):

ixTec (dismissed):
Approving but it would be great if we could also have a test case for this in `src/helpers/__tests__/set-title-case.spec.ts` too

ixTec (approved):
LGTM - thank you for the contribution!

PR #225, opened 2023-09-05
[WF-900] Theming integration for assorted base components

ixTec:
# Theming integration for Avatar, Badge, Chapeau, Code, Code Block, Skill Tag and Tag

### 🚀 Major Changes

- Integrate theming for the following components:
    - Avatar
    - Badge
    - Chapeau
        - Small inconsistency between `inverted` and themed colors, but this is fine as inverted isn't used much
    - Code
    - Code Block
    - Skill Tag
    - Tag
- Deprecated `inverted` prop for the above components

### 🔧 Minor Changes
- Fixed issue with naming of `transparentGrey` accent tokens (was `transparent-grey`)

[WF-900]: https://datacamp.atlassian.net/browse/WF-900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/avatar/styles.ts
```diff
@@ -107,27 +67,43 @@ export function avatarStyle({
...
 }: AvatarStyleOptions) {
-  return css`
-    // Fix for background-color and border-radius combo artifacts
-    background: ${contentFill
-      ? `radial-gradient(
-      farthest-side,
-      ${variantMap[variant].backgroundColor} calc(100% - 1px),
-      transparent)`
-      : variantMap[variant].backgroundColor};
+  const variantStyle =
```
dandenney, line 70: Is this a necessary override as part of the changes? I read it as deprecating `greyLight`, so just checking that that is the goal

dandenney: I'm guessing so since the test checks for it https://github.com/datacamp/waffles/blob/8e64a4f2ae754aee33c1d498b3be7c7e779df32d/src/badge/__tests__/__snapshots__/badge.spec.tsx.snap#L417

src/badge/styles.ts
```diff
@@ -21,87 +22,43 @@ const sizeMap = {
...
+  alignItems: 'center',
+  justifyContent: 'center',
+  fontFamily: tokens.fontFamilies.sansSerif,
+  fontWeight: tokens.fontWeights.bold,
+  lineHeight: tokens.lineHeights.tight,
+  textTransform: 'uppercase',
+  borderRadius: tokens.borderRadius.medium,
+  paddingLeft: tokens.spacing.xsmall,
+  paddingRight: tokens.spacing.xsmall,
+  letterSpacing: '0.5px',
```
dandenney, line 36: Non-blocker, but I wonder if we should add this to the tokens or change it to a tokenized value. I know the latter is a bigger decision. I don't know what a word is for between default and relaxed, but ChatGPT had some suggestions

> The term "casual" could serve as an intermediary between "default" and "relaxed." While "default" often refers to a standard or baseline setting, condition, or behavior, "relaxed" implies a more laid-back or easygoing state. "Casual" suggests a blend of these, denoting an informal or nonchalant approach, but not as unstructured or as leisurely as "relaxed."

> Depending on the context in which you're using these words, other terms like "informal," "moderate," or "easygoing" might also fit.

src/badge/styles.ts
```diff
@@ -21,87 +22,43 @@ const sizeMap = {
...
+      color: variantStyle.textOnColor,
+      backgroundColor: variantStyle.main,
+      fontSize: sizeMap[size].fontSize,
+      height: sizeMap[size].sizing,
+    }),
+  ];
 }
+
+export const innerContentStaticStyle = css({
+  maxWidth: '164px',
```
dandenney, line 60: I know it's old, but I've always wondered why the default had this. Feels like it should be a "trim" variable or something

dandenney (approved):
Approving because nothing jumps out to me outside of your documented Chapeau issues and the greyLight which I believe is intentional. Added my usual nits on style bits that are outside of the scope of this and are non-blocking. One can dream.

src/avatar/styles.ts
ixTec: Aye, `greyLight` is being replaced with `transparentGrey`

PR #232, opened 2023-09-12
[WF-902] Decorative component theming integration

ixTec:
# Integrating theming into our decorative components

### 🚀 Major Changes

- Integrate theming support and deprecate `inverted` prop for the following components:
    - Asset
    - Brand
    - Icon
    - Loader

### 🔧 Minor Changes

- Clean up `Button` internal structure to remove new wrapper

[WF-902]: https://datacamp.atlassian.net/browse/WF-902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/brand/product/workspace/workspace-logo.tsx
```diff
@@ -1,21 +1,17 @@
...
   ...restProps
 }: BrandProps) {
-  const textColor = inverted ? tokens.colors.white : tokens.colors.navy;
-  const logomarkColor = monochrome
-    ? inverted
-      ? tokens.colors.white
-      : tokens.colors.navy
-    : tokens.colors.green;
+  const textColor = themedTokens.text.main;
+  const logomarkColor = monochrome ? textColor : themedTokens.brand.main;
```
lensbart, line 11: I think we should inline `themedTokens.text.main` here and elsewhere

lensbart (approved):

src/brand/product/workspace/workspace-logo.tsx
ixTec: It should always be the same as `textColor` when monochrome, so I think it's better to use that value than have it (potentially) different if we change `textColor` in the future

lensbart: Oh, I missed that `textColor` was used elsewhere — my bad!

PR #235, opened 2023-09-20
[WF-903] Modal-based component theming integration

ixTec:
# Integrate theming for Alert Dialog, Dialog, Drawer, Popover and internal Modal component

### 🚀 Major Changes

- Integrate theming support for the following components:
    - Alert Dialog
    - Dialog
    - Drawer
    - Modal (internal component)
    - Popover

### 🔧 Minor Changes

- Slightly update `valueFromCssVariable` helper

### ❓ Misc Changes

- Fixed typo in Popover usage guidance
- Removed `new` badge from Popover in docs

[WF-903]: https://datacamp.atlassian.net/browse/WF-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/helpers/value-from-css-variable.ts
```diff
@@ -8,7 +8,16 @@ import { CssVariable } from './types';
...
+  target?: HTMLElement,
 ) {
-  return getComputedStyle(target).getPropertyValue(cssVariable) || null;
+  // If document isn't valid, then we cannot get the value of the CSS variable
+  if (typeof document === 'undefined') {
+    return null;
+  }
+
+  return (
+    getComputedStyle(target || document.documentElement).getPropertyValue(
```
lensbart, line 19: Since it’s a `null`/`undefined` check, the following feels a bit more precise:
```suggestion
    getComputedStyle(target ?? document.documentElement).getPropertyValue(
```
(and will be enforced by the DataCamp ESLint preset)

src/helpers/value-from-css-variable.ts
```diff
@@ -8,7 +8,16 @@ import { CssVariable } from './types';
...
+      cssVariable,
+    ) || null
```
lensbart, line 21:
```suggestion
    ) ?? null
```
Although I’m not sure if this really should be `null`? Anyway, it doesn’t matter much either way.

src/dialog/dialog-internal.tsx
```diff
@@ -70,9 +67,9 @@ function DialogInternal({
         alignCenter,
         ...restProps,
       }}
-      css={css`
-        position: relative;
-      `}
+      // css={css`
+      //   position: relative;
+      // `}
```
lensbart, line 72: Intentionally left in?

src/popover/styles.ts
```diff
@@ -28,64 +35,51 @@ type WrapperStyleOptions = {
...
-export function countStyle() {
-  return css`
-    display: flex;
-    color: ${tokens.colors.navySubtleTextOnLight};
-    line-height: ${tokens.lineHeights.relaxed};
-    flex-grow: 1;
-    user-select: none;
-  `;
-}
+export const countStaticStyle = css({
```
lensbart, line 88: I’m not sure I see the benefit of including `Static` in all of these constants:
* The type reveals that the constant is of type `SerializedStyles`, so that’s distinct from functions and should be sufficiently informative
* Some of these styles are responsive (such as `wrapperStaticStyle` above, using `mediaQuery`), so the current nomenclature might be confusing
* Brevity

lensbart (approved):
L🅰️➕TM

src/popover/styles.ts
ixTec: It's a naming convention I adopted when cleaning up the styling. It's not necessary, just helps distinguish what should/shouldn't be dynamic CSS.

Will leave it as is, but if it's too much in the future then it may be reviewed again

PR #236, opened 2023-09-22
[WF-903] Theming integration for Empty State, Tabs and Tooltip

ixTec:
# Theming integration for Empty State, Tabs and Tooltip

### 🚀 Major Changes

- Integrate theming support and deprecate the `inverted` prop for the following components:
    - Empty State
    - Tabs
    - Tooltip

### ❓ Misc Changes

- Update missed JSDoc comments

[WF-903]: https://datacamp.atlassian.net/browse/WF-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/empty-state/styles.ts
```diff
@@ -26,81 +27,83 @@ export function emptyStateStyle({
...
 export function imageStyle({ hasSmallWidth }: ImageStyleOptions) {
-  return css`
-    flex-shrink: 0;
-    width: ${hasSmallWidth ? 72 : 100}px;
-    height: ${hasSmallWidth ? 72 : 100}px;
-  `;
+  return [
+    imageStaticStyle,
+    css({
+      width: hasSmallWidth ? 72 : 100,
```
Rahien, line 60: ❓ worth turning into a token you think?

src/table/styles.ts
```diff
@@ -305,7 +305,7 @@ const cellMenuTriggerStaticStyle = css({
   padding: `0 ${tokens.spacing.small}`,
   height: tokens.sizing.small,
   minWidth: tokens.sizing.small,
-  marginLeft: -tokens.spacing.small,
```
Rahien, line 308: nice fix!

src/tooltip/styles.ts
```diff
@@ -1,15 +1,16 @@
...
-  position: absolute; // Use absolute position, expected by float-ui default strategy
-  z-index: 10;
-  padding: ${tokens.spacing.small} 12px;
-  border-radius: ${tokens.borderRadius.medium};
-  line-height: ${tokens.lineHeights.default};
-  max-width: 300px;
-`;
+const tooltipStaticStyle = css({
+  position: 'absolute', // Use absolute position, expected by floating-ui default strategy
+  zIndex: 10,
```
Rahien, line 8: ❓ so i've used this one as an in between zIndex before too... is it useful to name these as tokens?

Rahien (approved):

src/tooltip/styles.ts
ixTec: I think as it's an exceptional case here, it's better to leave it as a hardcoded thing - we don't want to encourage the use outside 😄

src/empty-state/styles.ts
ixTec: I think it's use case is a bit limited, but where we review the component as planned in a future quarter then it will probably get changed anyway

